PR #273, opened 2023-05-31
feat: tokens pagination info

### LINKED ISSUE #33: Tokens counter
The endpoint is not existent yet. Once implemented we can add the counter UI to tokens page

Tracking https://github.com/aeternity/ae_mdw/issues/1255

lukeromanowicz:
## Description
Resolves #33 

Implements tokens pagination info on the tokens page.

## Demo
image
image

## Checklist:


src/components/TokensPanel.vue
```diff
@@ -43,6 +44,11 @@ async function loadNextTokens() {
 
 const limit = computed(() => process.client && isDesktop() ? 10 : 3)
 
+await useAsyncData(async() => {
```
janmichek, line 47: Other usages of `useAsyncData` does not have to return true. EG `fetchOracleDetails`. Why here is it different?

src/stores/tokens.js
```diff
@@ -8,10 +8,14 @@ export const useTokensStore = defineStore('tokens', () => {
   const rawListedTokens = ref(null)
   const allTokens = ref(null)
   const selectedTokenName = ref(null)
+  const allTokensCount = ref(null)
```
janmichek, line 11: Personally, I am not really a fan of using `all`. We have not used that for other entities. I think it should explain something else. Could it make sense without using `all` ?

janmichek (changes requested):
The functionality is superb 🍰 Take a look at my code comments

src/components/TokensPanel.vue
lukeromanowicz: They always do (or at least they should for ensuring correct deduplication of requests between SSR and CSR). In the fetchOracleDetails return true is included in the fetch function, but all of our simple fetch functions return Promise<void> so I just added return true in the useAsyncData to keep it consistent. Proof:

image

src/stores/tokens.js
lukeromanowicz: @janmichek I used `all` because it refers directly to the entities stored in allTokens ref that you have originally implemented 2 lines above. I assume you chose that prefix to distinguish allTokens related things (unfiltered) from filtered things like rawListedTokens.

Alternative naming that comes to my mind is unfilteredTokens* but I like allTokens* more. Do you have any nice suggestions how to improve the naming?

src/components/TokensPanel.vue
janmichek: Understood. Thanks for explanation

src/stores/tokens.js
janmichek: Oh that was me 😅  I will try to come up with something...

michele-franchi (changes requested):
Pagination count does not look accurate, could you please have a look at it?
image

src/stores/tokens.js
janmichek: I went through the code again and remember that the inconsistent and annoying `all` prefix is used on purpose as the data comes from different sources. Here make sense. Thanks for refreshing my initial thinking. No need for change

janmichek (approved):

lukeromanowicz:
Thanks Michele, I didn't notice that. At a first glance it looks like a middleware issue.

Reported in https://github.com/aeternity/ae_mdw/issues/1372

lukeromanowicz:
The issue has been resolved on the mdw side.

lukeromanowicz:
Thanks @CharisSiarampalis, I missed it. Should be fixed now.

michele-franchi (approved):
It looks great now! Thanks for fixing it 👏

CharisSiarampalis:
@lukeromanowicz Issue with the records was fixed.
But if you go to the last page :
image
It says 431-440 out of 450.
Either the total number is wrong or there should be another page.

lukeromanowicz:
@CharisSiarampalis another very good catch. Should be working correctly now

CharisSiarampalis:
@lukeromanowicz On desktop it seems ok.
But on mobile the records are broken.
Only the first three out of ten are being displayed. Or maybe its just what the record says.
Also the record goes beyond the maximum number.


https://github.com/aeternity/aescan/assets/42606212/e561c360-1334-46fe-824d-36d2364e6ef8

lukeromanowicz:
@CharisSiarampalis right, I forgot to test in on mobile again after recent changes. I'm pretty sure now it should work :+1:

CharisSiarampalis (approved):

PR #287, opened 2023-06-05
feat: Enable state channel counter 

### LINKED ISSUE #233: Enable state channel counter
This is a followup of https://github.com/aeternity/aescan/pull/206

The count endpoint was returning different count
https://mainnet.aeternity.io/mdw/v2/txs/count?tx_type=channel_create (119)

than after manual counting state channels from listing (102)
https://mainnet.aeternity.io/mdw/v2/channels?limit=100

because the listing response does not include closed channels

TRACKING https://github.com/aeternity/ae_mdw/issues/1320

Same count from count endpoint and what we get from listing endpoint

No

janmichek:
## Description
fixes #233 
now went the MDW issue has been resolved https://github.com/aeternity/ae_mdw/issues/1320 (adding closed state channels to listing), the count endpoint will be in sync with the amount in the list.

## Demo

https://github.com/aeternity/aescan/assets/15363559/933eaef9-64c1-4303-ae2e-a3a3146ccee0

## Checklist:

- [x] re-enable state channels counter

### DIFF
--- a/src/components/StateChannelsPanel.vue
+++ b/src/components/StateChannelsPanel.vue
@@ -48,9 +48,10 @@ const loadNextStateChannels = () => {
 
 const loadStateChannels = () => {
   fetchStateChannels({ limit: limit.value })
-  fetchStateChannelsCount()
   pageIndex.value = 1
 }
+
+fetchStateChannelsCount()
 
 if (process.client) {
   loadStateChannels()
--- a/src/components/StateChannelsPanel.vue
+++ b/src/components/StateChannelsPanel.vue
@@ -6,6 +6,7 @@
       v-model:page-index="pageIndex"
       :entities="stateChannels"
       pagination-style="history"
+      :total-count="stateChannelsCount"
       :limit="limit"
       @prev-clicked="loadPrevStateChannels"
       @next-clicked="loadNextStateChannels">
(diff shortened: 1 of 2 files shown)


src/stores/stateChannels.js
```diff
@@ -31,7 +31,14 @@ export const useStateChannelsStore = defineStore('stateChannels', () => {
     rawStateChannels.value = data
   }
 
+  async function fetchStateChannelsCount() {
+    const { data } = await axios.get(`${useRuntimeConfig().public.MIDDLEWARE_URL}/v2/txs/count?tx_type=channel_create`)
```
lukeromanowicz, line 35: how about we move useRuntimeConfig().public to the top of the store definition and leave here only the MIDDLEWARE_URL?

src/components/StateChannelsPanel.vue
```diff
@@ -47,6 +48,7 @@ const loadNextStateChannels = () => {
 
 const loadStateChannels = () => {
   fetchStateChannels({ limit: limit.value })
+  fetchStateChannelsCount()
```
lukeromanowicz, line 51: I guess this could be fetched on the server-side but it's not a big deal. I'm fine with leaving it as it is

lukeromanowicz (approved):
Nice job, I approve it in its current state, Comments below are optional to fix

src/stores/stateChannels.js
janmichek: Right, it was already pre-defined in file. Fixed

src/components/StateChannelsPanel.vue
janmichek: Yep, let's keep it consistent. FIxed

michele-franchi (approved):
Looks great 👍

CharisSiarampalis (approved):

Issue #198, opened 2023-05-03
Improve blockchain panel implementation


lukeromanowicz:
### Please describe the problem that your request should resolve.
Once https://github.com/aeternity/ae_mdw/issues/1291 is resolved we should adjust our blockchain panel implementation accordingly to improve the data accuracy and get rid of data inconsistencies.

michele-franchi:
@janmichek to check the history of this and see if it's actually possible to get rid of the multiple HTTP requests that are being made and switch to WebSocket.

michele-franchi:
@janmichek to add comments about the current status

janmichek:
> @janmichek to add comments about the current status

Waiting for response from MDW https://github.com/aeternity/ae_mdw/pull/1382#issuecomment-1808381950

michele-franchi:
@janmichek to check again the status of this issue

PR #295, opened 2023-06-07
feat: revert Blockchain Panel to rely on REST API

### LINKED ISSUE #292: Temporarily use REST API for data fetching in blockchain section
Temporarily use REST API for data fetching in blockchain section until the websocket-based solution discussed in #198 is possible

lukeromanowicz:
## Description
Resolves #292 

## Checklist:

### DIFF
--- a/src/stores/recentBlocks.js
+++ b/src/stores/recentBlocks.js
@@ -4,13 +4,13 @@ import { useRuntimeConfig } from 'nuxt/app'
 import { useBlockchainStatsStore } from '@/stores/blockchainStats'
 import { adaptDeltaStats, adaptKeyblock, adaptSelectedMicroblockTransactions } from '@/utils/adapters'
 import { formatAettosToAe, formatNullable, formatNumber } from '@/utils/format'
-import { MESSAGE_BUFFER_SIZE_LIMIT, VISIBLE_KEYBLOCKS_LIMIT, VISIBLE_MICROBLOCKS_LIMIT, VISIBLE_TRANSACTIONS_LIMIT } from '@/utils/constants'
+import { VISIBLE_KEYBLOCKS_LIMIT, VISIBLE_MICROBLOCKS_LIMIT, VISIBLE_TRANSACTIONS_LIMIT } from '@/utils/constants'
 
 const isBlockFirstInSequence = (block, blockSequence) => block.hash === blockSequence?.[0].hash
 
 export const useRecentBlocksStore = defineStore('recentBlocks', () => {
   const { MIDDLEWARE_URL } = useRuntimeConfig().public
-  const { fetchTotalTransactionsCount, increaseTransactionsCounter } = useBlockchainStatsStore()
+  const { fetchTotalTransactionsCount } = useBlockchainStatsStore()
 
   const deltaStats = ref(null)
   const keyblocks = ref(null)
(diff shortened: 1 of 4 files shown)


janmichek:
Question to this one: So the batch websocket update is not gonnabe resolved by this? It still depends on MDW fix, right?

janmichek (changes requested):
Sometimes I noticed 
`BRI-Reward` and `Beneficiary-Reward`  is not being loaded. Just `---`. COuld you check please?

lukeromanowicz:
@janmichek it was a temporary middleware issue. Unrelated to the code. Now it's working correctly

CharisSiarampalis (approved):

janmichek:
2 more questions @lukeromanowicz 
1) This is the reverted previous code, are there new ideas introduced?
2) Do we have a followup ticket for 'proper' solution?

lukeromanowicz:
@janmichek 
1. previous code was not using composition API and data fetching triggered every time there was a new transaction message from the websocket. I ported the exact same logic that was used with the only change: the data fetch happens only to every time there is a new microblock minted because microblock and transaction updates happen together at the very same time.
2. we do: https://github.com/aeternity/aescan/issues/198

janmichek (changes requested):
There is still an issue with reward values.
Scenario:
1. wait for new block being minted.
2. 2 reward values will change to ---
3. when you refresh it will become actual value


https://github.com/aeternity/aescan/assets/15363559/c0825595-fd1a-4f3a-8c94-74748af187ff

lukeromanowicz:
@janmichek thanks Jan, good eye. Fixed! The issue was caused by the delay in processing the new keyblock data between the node websocket and the REST API

https://github.com/aeternity/aescan/assets/46789227/4c34878c-048a-4f45-948f-48dfccffece0

src/stores/recentBlocks.js
```diff
@@ -143,160 +127,40 @@ export const useRecentBlocksStore = defineStore('recentBlocks', () => {
...
+        await fetchSelectedKeyblockMicroblocks(selectedKeyblock.value.hash)
+
+        if (isFirstMicroblockSelected.value && isFirstKeyblockSelected.value) {
+          await fetchSelectedMicroblockTransactions()
+        }
+      } catch (error) {
+        // ignore 400 errors when fetching data by a non-existing microblock
+        // as they are caused by microforks
+        if (error?.response.status !== 400) {
+          throw error
```
michele-franchi, line 149: Is this handled somewhere?

src/stores/recentBlocks.js
```diff
@@ -143,160 +127,40 @@ export const useRecentBlocksStore = defineStore('recentBlocks', () => {
...
+        }
       }
 
-      if (success) {
-        messageBuffer.value.splice(i, 1)
-        return true
+      // sometimes delta stats are not yet available on keyblock message, so retry fetching them again
+      if (selectedDeltaStats.value === null) {
+        await fetchDeltaStats()
```
michele-franchi, line 155: Is this a temporary solution (mdw issue)? If yes, is there a task to improve it?

michele-franchi (approved):
I left some comments but other than that it looks really good. Well done!

src/stores/recentBlocks.js
lukeromanowicz: No, just like it wasn't before. I don't expect this to ever happen but if it does we will receive a notification via Sentry

janmichek (approved):
Looks great 🍻  thanks for the fix

PR #300, opened 2023-06-09
fix: support for internal call event in name actions

### LINKED ISSUE #267: Name history does not work for some names
### Context information

application version: 
device: 
browser: 
operating system: 

### Steps to reproduce
- Open https://aescan.io/names/yani.chain
- Check console

### What is expected?
Name history is displayed correctly.

image

lukeromanowicz:
## Description
Resolves #267 

The problem was the lack of support for internal contract call events which have a different payload.

## Demo
image

## Checklist:

### DIFF
--- a/src/stores/nameDetails.js
+++ b/src/stores/nameDetails.js
@@ -18,7 +18,7 @@ export const useNameDetailsStore = defineStore('nameDetails', () => {
     return rawName.value?.hash || rawName.value?.info.last_bid.tx.name_id
   })
   const nameActions = computed(() => {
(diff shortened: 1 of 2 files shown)


janmichek (approved):
Cool, it works now. Also thanks for the name refactoring. 🦖 
Was this working before or it was never working? I haven't spotted any issue before, but I am not sure

lukeromanowicz:
@janmichek I think that same names started using events that we have never been supporting before

CharisSiarampalis (approved):

PR #317, opened 2023-06-14
fix: Name Tabs linking regression

### LINKED ISSUE #278: Name Tabs linking regression
### Context information

all platforms
https://aescan.dev.aepps.com/

### Steps to reproduce

https://github.com/aeternity/aescan/assets/15363559/d8929356-772f-438d-b48b-74b0171738ae

### What is expected?
The URL will change after selecting tab

URL is not changed

janmichek:
## Description
fixes #278 

## Demo

https://github.com/aeternity/aescan/assets/15363559/319457cf-8492-4fa3-81cd-e3e3a0d147e6

## Checklist:

- [x] fix tabs linking

### DIFF
--- a/src/components/AppTabs.vue
+++ b/src/components/AppTabs.vue
@@ -26,7 +26,7 @@ const props = defineProps({
   },
 })
 
-const emit = defineEmits(['update:model-value'])
+const emit = defineEmits(['update:modelValue'])
 
 const activeTabIndex = props.modelValue ? useVModel(props, 'modelValue', emit) : ref(0)
 const tabs = ref([])


lukeromanowicz (approved):
Thanks for fixing it. I have no idea how I could missed it when working on this :/

CharisSiarampalis (approved):

michele-franchi (changes requested):
For some reason it does not work when you click on "Show all" in the dashboard panel "Names recently activated". If you try to switch tabs the value in the url does not change.

I also think the query param should be appended when you directly visit the names page.

janmichek:
> For some reason it does not work when you click on "Show all" in the dashboard panel "Names recently activated". If you try to switch tabs the value in the url does not change.
> 


For me it works as expected


https://github.com/aeternity/aescan/assets/15363559/36704f3a-31b8-42fc-9486-84e4b72fe9c5




> I also think the query param should be appended when you directly visit the names page.

Could be... not necessarily. It is how it was working before

michele-franchi:
@janmichek regarding my comment:
> For some reason it does not work when you click on "Show all" in the dashboard panel "Names recently activated". If you try to switch tabs the value in the url does not change.

The key part is:
> If you try to switch tabs the value in the url does not change.

But if you follow the same procedure with "Auctions ending soon" the url changes when you change tabs so it looks like something is broken.

janmichek:
Aha I see now. Will fix it

janmichek:
Fixed, demo regenerated

CharisSiarampalis (approved):

michele-franchi (changes requested):
There's still an issue if you directly navigate to the names page from the navigation bar. The query param `?type=active` gets appended and you have to navigate back two times to go to the dashboard (or the previous page).
I would fix this behaviour or totally avoid adding the query param if you simply navigate to the /names page.

janmichek:
I see, apparently, that's how was the behavior even before refactoring 
https://aescan.io/names

In my eyes, it's not really nice code. Overall complicated and there is some confusing naming. This part would deserve rewriting.

janmichek:
I was struggling with this a lot, I would prefer to postpone it

lukeromanowicz:
Fixed in #351

PR #358, opened 2023-06-28
fix: Oracle response of OracleRespondTx transactions is not displayed correctly

janmichek:
## Description
This pr has been forgotten to merge in previous MVP private repo, so just moving it here

## Demo
Before 
image

After
image

## Checklist:
- [x] [truncate Oracle respond data in diffrent way

### DIFF
--- a/src/components/TransactionCellOracleRespondTx.vue
+++ b/src/components/TransactionCellOracleRespondTx.vue
@@ -1,5 +1,8 @@
 <template>
-  <value-hash-ellipsed :hash="transactionData.oracle_id"/>
+  <value-hash-ellipsed
+    :hash="transactionData.oracle_id"
+    :link-to="`/oracles/${transactionData.oracle_id}`"/>
+
   <transaction-arrow-right-icon/>
 
   <value-hash-ellipsed :hash="transactionData.query_id"/>
--- a/src/components/TransactionCellOracleRespondTx.vue
+++ b/src/components/TransactionCellOracleRespondTx.vue
@@ -7,7 +7,9 @@
 
   <value-hash-ellipsed :hash="transactionData.query_id"/>
 
-  <app-chip size="sm">
+  <app-chip
+    v-if="oracleResponse"
+    size="sm">
     {{ oracleResponse }}
   </app-chip>
 </template>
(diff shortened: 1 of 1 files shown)


src/components/TransactionCellOracleRespondTx.vue
```diff
@@ -1,8 +1,5 @@
 <template>
-  <value-hash-ellipsed
```
lukeromanowicz, line 2: Shouldn't we keep the oracle url?

janmichek: Yes we should, fixed

lukeromanowicz (approved):
Not sure if we should show the empty chip. Maybe we should just hide it if the response is empty.

Nonetheless, it's a good improvement - thanks!

janmichek:
@michele-franchi Waiting for review from you

michele-franchi (changes requested):
I had a look at the original issue and I think it's still not full solved.

As you can see from the screenshot, empty values are printed:
image

What do you think about hiding them?

In general what we discussed here back then was to find a way to get rid of the responses that are for example:
- []...[]
- ye...es
- 8...8

And If I remember correctly my suggestion was the following:
- Hide the value if it's empty
- Print the value when it's under 8 chars
- Use the hash value component if it's more than 8 chars


I also noticed, not really related to your fix, that some responses are not printed at all. As you can see in the screenshot the first rows should contain "good":
image

Feel free to open a separate issue if you think this is not the right place.

janmichek:
I successfully resolved the ellipsing of short responses.

The other problem with `no response` is a bit more complicated.
The problem is that some responses are encoded and some are note.

I am not really sure if it's an issue as this appears only on testnet? Maybe It was a product of some experimentation? Really not sure here

src/components/TransactionCellOracleRespondTx.vue
```diff
@@ -16,15 +18,16 @@
...
   transactionData: {
     required: true,
     type: Object,
   },
 })
-const oracleResponse = computed(() =>
-  formatEllipseHash(formatDecodeByteArray(props.transactionData.response)),
-)
+const oracleResponse = computed(() => {
+  const decodedResponse = formatDecodeByteArray(props.transactionData.response)
```
michele-franchi, line 30:
```suggestion
  const { response } = props.transactionData
  const decodedResponse = response instanceof Array ? formatDecodeByteArray(response) : response.toString()
```

This should also handle empty cases, what do you think?

michele-franchi (commented):
I added a suggestion for the empty cases, I tested locally with testnet and it works but feel free to adjust the code if you don't like it 😄

src/components/TransactionCellOracleRespondTx.vue
janmichek: Ok, good. I wanted to avoid fixing data inconsistencies on FE, but ok, I accept your suggestion. It's a bit better solution
However, it didn't solve it ultimately. There are still some characters that cannot be printed, looking like it's 'empty'

image

michele-franchi (approved):
Looks good now, great job 👏

Issue #375, opened 2023-07-14
Implement switching to next/previous block on detail pages


lukeromanowicz:
### Please describe the problem that your request should resolve.
Implement navigation to previous/next keyblock and microblock on the respective details pages

PR #376, opened 2023-07-14
feat: keyblock details page

### LINKED ISSUE #225: Keyblock details page
A new panel should be added with the following:
- Keyblock Hash
- Height
- Time
- Miner
- Beneficiary
- Reward
- Microblocks Count
- Transactions Count

Something should also be added to tell the user there's a sequence in the keyblocks and it would be nice to navigate between them.

lukeromanowicz:
## Description
Resolves #225 

Navigation to the previous and next keyblock has been split into #375 since there is still some open discussion on how it is supposed to work.

## Demo
image

image

## Checklist:


lukeromanowicz:
Example: https://pr-376-aescan.stg.aepps.com/keyblocks/799897 or kh_MGnv8gVCSm4RprcrqMTgSMCoRPFYFNFSxERZG7Y8vrejcWExF

src/components/KeyblockDetailsPanel.vue
```diff
@@ -0,0 +1,220 @@
...
+          :label="formatEllipseHash(keyblockDetails.hash)"
+          :clipboard-text="keyblockDetails.hash"
+          class="keyblock-details-panel__chip-ellipse"/>
+      </div>
+    </header>
+    <table>
+      <tbody>
+        <tr class="keyblock-details-panel__row">
+          <th class="keyblock-details-panel__table-header">
+            Keyblock height
```
janmichek, line 21: As for the unification of hints (and labels), we agreed that all labels should be capitalized. The content of hints should not be capitalized
```suggestion
            Keyblock Height
```

janmichek (changes requested):
Looks decent so far 🦾 
Please look at my code comments.

Feature-wise this is something I don't understand:
When I click on the Miner detail, It should be existent, or no?

https://github.com/aeternity/aescan/assets/15363559/0fc3131c-7544-4bc7-b864-4ba1c502957f

src/utils/adapters.js
```diff
@@ -8,11 +8,13 @@ function isAuction(chainName) {
   return chainName.length - suffixLength < auctionLength
 }
 
-export function adaptKeyblock(keyblock) {
+export function adaptKeyblock(keyblock, keyblockDeltaStats = null) {
   if (keyblock) {
     return {
       ...keyblock,
       mined: DateTime.fromMillis(keyblock.time),
+      block_reward: keyblockDeltaStats ? formatAettosToAe(keyblockDeltaStats.block_reward) : null,
```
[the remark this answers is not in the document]
lukeromanowicz: It's not sanitized in the argument - it's only hinting that keyblockDeltaStats is optional. Depending on keyblockDeltaStats being defined or not block_rewards are added or not to the returned value

src/utils/adapters.js
```diff
@@ -8,11 +8,13 @@ function isAuction(chainName) {
   return chainName.length - suffixLength < auctionLength
 }
 
-export function adaptKeyblock(keyblock) {
+export function adaptKeyblock(keyblock, keyblockDeltaStats = null) {
   if (keyblock) {
```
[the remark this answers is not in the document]
lukeromanowicz: Not sure why we originally were checking if the required argument is passed or not. We usually don't do this in other places. I just left it unchanged not to break something

src/components/KeyblockDetailsPanel.vue
```diff
@@ -0,0 +1,220 @@
...
+const { NODE_URL, MIDDLEWARE_URL } = useRuntimeConfig().public
+
+const props = defineProps({
+  keyblockDetails: {
+    type: Object,
+    required: true,
+  },
+})
+
+const keyblockNodeUrl = computed(() =>
```
janmichek, line 144: One more thing to consider - Would this find a better home in adapters? It is also valid for more entities. Maybe it is worth refactoring issue?

lukeromanowicz: Interesting idea. I like keeping it in one place along with other data transformation but on the other hand it's used only in one place per entity type while the adapters are generally reusable (e.g. adaptKeyblock is used in multiple places). I don't have a strong opinion. @michele-franchi wdyt?

janmichek:
@lukeromanowicz I think this has not been answered

> When I click on the Miner detail, It should be existent, or no?

marc0olo:
@janmichek the miner keypair is rotated frequently afaik and typically doesn't receive funds. funds are transferred to the beneficiary account defined by the miner. so yeah, account not found will almost anytime be true in case of the miner 🤔 

can somebody confirm? @velzevur @seanhinde @uwiger @hanssv

hanssv:
Yes, confirmed. The keypair used to sign the blocks is throw-away and only used for a single generation(*) - the reward goes to the the beneficiary. 

(*) in the only existing implementation - there is nothing in the protocol dictating this. However, it would be bad if you managed to mine a key-block using a leaked key-pair (someone else could provide valid microblocks!), so a throw-away keypair makes sense.

janmichek:
> Yes, confirmed. The keypair used to sign the blocks is throw-away and only used for a single generation(*) - the reward goes to the the beneficiary.
> 
> (*) in the only existing implementation - there is nothing in the protocol dictating this. However, it would be bad if you managed to mine a key-block using a leaked key-pair (someone else could provide valid microblocks!), so a throw-away keypair makes sense.

Cool, thanks for extensive explanation

src/components/KeyblockDetailsPanel.vue
```diff
@@ -0,0 +1,220 @@
...
+          :label="formatEllipseHash(keyblockDetails.hash)"
+          :clipboard-text="keyblockDetails.hash"
+          class="keyblock-details-panel__chip-ellipse"/>
+      </div>
+    </header>
+    <table>
+      <tbody>
+        <tr class="keyblock-details-panel__row">
+          <th class="keyblock-details-panel__table-header">
+            Keyblock Height
```
michele-franchi, line 21:
```suggestion
            Height
```

I would remove `Keyblock`, I see it as a repetition

michele-franchi (approved):
Good job! I left a suggestion for a row name.

src/components/KeyblockDetailsPanel.vue
janmichek: Yes that's my suggestion. My intention to create adapters was to move as most logic from templates and create the layer that would move mapping from MDW aside

src/components/KeyblockDetailsPanel.vue
lukeromanowicz: Thank you, fixed

src/utils/adapters.js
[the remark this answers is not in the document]
janmichek: I am not really satisfied with the sudden change of casing without further discussion. There are more ways how to make it consistent. Now the code casing is inconsistent. It may take a long time till we will start using typescript so I don't like the idea. I suggest making it back consistent in the open PR's or creating refactoring issue to make it consistent real soon.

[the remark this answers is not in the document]
lukeromanowicz: @janmichek I disagree this is a change or introducing more inconsistency. These properties are called this way in our application for many months:

image

I'm all for unifying everything to one case but we need to do it everywhere, not selectively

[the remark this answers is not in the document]
lukeromanowicz: I created a dedicated task to unify that: https://github.com/aeternity/aescan/issues/385

PR #380, opened 2023-07-18
feat: keyblock and microblock links

### LINKED ISSUE #366: Make keyblocks and microblocks clickable
Keyblocks should be clickable (hash and height):
image
image

Microblocks should be clickable:
image

lukeromanowicz:
## Description
Resolves #366 
Implements clickable keyblock and microblock hashes/block heights all over the app.
Keyblocks that have not been minted yet should NOT be links.

To be merged AFTER keyblock and microblock details pages. Currently the links lead to nowhere until the previous tasks get merged

## Demo
image

## Checklist:


src/components/ContractsTable.vue
```diff
@@ -9,9 +9,9 @@
           </hint-tooltip>
         </th>
         <th>
-          Created By
+          Created
```
lukeromanowicz, line 12: Unrelated, I changed it from "Created By" back to "Created" because on current develop "Created By" header is duplicated in two columns.

src/utils/adapters.js
```diff
@@ -407,7 +407,7 @@ export function adaptOracleDetails(oracle, lastExtendedTx, lastQueryTx, blockHei
   const oracleDetails = {
     id: oracle.oracle,
     fee: formatAettosToAe(oracle.query_fee),
-    expiration: oracle.approximate_expire_time,
+    expiration: DateTime.fromMillis(oracle.approximate_expire_time),
```
lukeromanowicz, line 410: Unrelated: fix for some regression introduced in previously merged PRs

CharisSiarampalis (approved):

janmichek (approved):
Looks great, ser! 🚀 

I am giving 2 suggestions to consider

1) Is it right to have both routes to keyblock detail? `/keyblocks/kh_2a9D6kQcWAtK6AzLAivi38BeFbcVUF4Tcp25RPvrWqNoUoQWHj` and `/keyblocks/801118` ? Shouldn't this be translated somehow to just one?
In previous explorer was defined with just sequenced number.
There is param `keyblockHash` is keyblocksDetailsStore, which sometimes is not a hash.

2) So there is no link to the currently mined keyblock. Wouldn't be more explainable UX to redirect to the currently mined keyblock detail with the corresponding message on it?

lukeromanowicz:
Thanks for the review, @janmichek. To answer your points:

1. It can be easily done. I followed my guts to support both similarly to the API, but we can do the redirects to use keyblock hash. @michele-franchi please define which behavior should be implemented here
2.  Not sure if I understand correctly. Can you, please, mark on a screenshot what part of the UI you are referring to?

janmichek:
> 2\. Not sure if I understand correctly. Can you, please, mark on a screenshot what part of the UI you are referring to?

I was not precise with the description. The thing I propose is to make all links to keyblock detail active. In the case of future ones, just display 'This keyblock has not been mined yet.'

lukeromanowicz:
Oh, interesting idea. I like it :+1: @michele-franchi do you approve? If you do, should I use the same design as for the accounts that don't have any activity yet?

michele-franchi:
> Looks great, ser! 🚀
> 
> I am giving 2 suggestions to consider
> 
> 1. Is it right to have both routes to keyblock detail? `/keyblocks/kh_2a9D6kQcWAtK6AzLAivi38BeFbcVUF4Tcp25RPvrWqNoUoQWHj` and `/keyblocks/801118` ? Shouldn't this be translated somehow to just one?
>    In previous explorer was defined with just sequenced number.
>    There is param `keyblockHash` is keyblocksDetailsStore, which sometimes is not a hash.
> 2. So there is no link to the currently mined keyblock. Wouldn't be more explainable UX to redirect to the currently mined keyblock detail with the corresponding message on it?

1. If it's possible to always use the keyblock sequence than we could to that but I don't see any issue with using both of them, it could be handy. If it's an issue we could start with the sequence and add also hash support later.
The only thing that comes to my mind is that we would have to fetch the keyblock height in the search bar if someone enters the hash.

2. Yes, that would be great. We could use the message we already have with the account.

janmichek:
1. if we plan to use the solution for both identifiers, then the param name `keyblockHash` is not correct

michele-franchi (changes requested):
Looks good but some parts are missing:

image

image

image

image

image

Could you please adjust them? Also please have a look at my comment above.

janmichek:
@michele-franchi That's the future ones I was mentioning above

lukeromanowicz:
@michele-franchi as stated in the description it was correct and expected up until our most recent discussion. I implemented the support for non-minted keyblocks so from now on all keyblock heights should be covered.

@CharisSiarampalis I did a lot of changes - could you, please, help us out, please, to make sure all block heights are links now (even those which are not minted yet)

CharisSiarampalis:
@lukeromanowicz Keyblock height is not clickable on mobile. Is this normal?

image

lukeromanowicz:
@CharisSiarampalis it wasn't. Thanks for reporting - fixed

CharisSiarampalis (approved):

michele-franchi (approved):
Good job 👏

PR #382, opened 2023-07-18
feat: Keyblock microblocks table

### LINKED ISSUE #369: Keyblock microblocks table
In the keyblock details page here should also be a table with the microblocks of the keyblock with the following columns:

- Microblock hash
- Time
- Transactions count

janmichek:
## Description
resolves #369

## Demo

https://github.com/aeternity/aescan/assets/15363559/c48725cb-3587-4aac-bf95-5c83289178bc

## Checklist:

- [x] list block's mircroblocks


src/components/KeyblockMicroblocksTable.vue
```diff
@@ -0,0 +1,50 @@
...
+        <hint-tooltip>
+          {{ keyblocksHints.transactionsCount }}
+        </hint-tooltip>
+      </th>
+    </tr>
+    <tr
+      v-for="microblock in microblocks.data"
+      :key="microblock.hash">
+      <td>
+        <app-link :to="`/micro-blocks/${microblock.hash}`">
```
michele-franchi, line 27:
```suggestion
        <app-link :to="`/microblocks/${microblock.hash}`">
```

The link is incorrect

michele-franchi (changes requested):
Good job, I have some comments.

Any reason why the hash is not truncated? It would be the only table like this.
image

It's not clear that in general these are microblocks, it is not written anywhere.

janmichek:
@michele-franchi 

>Any reason why the hash is not truncated? It would be the only table like this.

I was following the visual material. No strong preference here


>It's not clear that in general these are microblocks, it is not written anywhere.

I was also following the visual material. Here I would prefer label as well.

What do you think about those changes @danilosierrac ?

src/components/KeyblockMicroblocksTable.vue
janmichek: fixed

src/components/KeyblockMicroblocksPanel.vue
```diff
@@ -0,0 +1,69 @@
...
+import PaginatedContent from '@/components/PaginatedContent'
+import KeyblockMicroblocksTable from '@/components/KeyblockMicroblocksTable'
+import KeyblockMicroblocksTableCondensed from '@/components/KeyblockMicroblocksTableCondensed'
+import { isDesktop } from '@/utils/screen'
+
+const { keyblockMicroblocks: microblocks, keyblockDetails } = storeToRefs(useKeyblockDetailsStore())
+const { fetchKeyblockMicroblocks } = useKeyblockDetailsStore()
+const route = useRoute()
+const limit = computed(() => process.client && isDesktop() ? 10 : 3)
+fetchKeyblockMicroblocks({
```
lukeromanowicz, line 35: I think this should be fetched on the client side only. Right now it's fetched also on server-side but the SSR doesn't wait for the response so you don't see any warnings

lukeromanowicz (approved):
Looks really good. Thank you for implementing it! I left one small note but I approve in advance

michele-franchi (changes requested):
Please have a look at the recent changes I suggested in the design:
- A new tab is visible called "Microblocks"
I would still use truncated hashes for consistency but I don't have a strong preference.

src/components/KeyblockMicroblocksPanel.vue
janmichek: Thanks, fixed

janmichek:
> * A new tab is visible called "Microblocks"

Ok, both adjusted

michele-franchi (approved):
Good job, thanks for the fixes 👏

CharisSiarampalis (approved):

Issue #148, opened 2023-04-20
Make pagination text clickable


michele-franchi:
### Please describe the problem that your request should resolve.
Only pagination buttons are clickable. 

### Describe the solution you'd like

Make pagination text clickable.

### Is it already possible to achieve the same outcome in any other way? If so, how?
No

PR #395, opened 2023-07-24
feat: Implement switching to next/previous block on detail pages

### LINKED ISSUE #375: Implement switching to next/previous block on detail pages

### LINKED ISSUE #406: Microblocks tab is visible when the keyblock is not mined
### Context information

application version: 
device: 
browser: 
operating system: 

### Steps to reproduce
- Open a keyblock which is not mined

### What is expected?
The microblocks tab should be hidden.

image

janmichek:
## Description
resolves #375 
fixes #406

compared to the visual material, the design is old for consistency and will be improved here https://github.com/aeternity/aescan/issues/148

## Demo

https://github.com/aeternity/aescan/assets/15363559/ad317fd1-d667-4db6-8e55-e6f538238081

## Checklist:

- [x] add navigation through keyblock detail

### DIFF
--- a/src/components/KeyblockDetailsPanel.vue
+++ b/src/components/KeyblockDetailsPanel.vue
@@ -255,10 +255,6 @@ const isNextKeyblockMined = computed(() =>
     }
   }
 
-  &__keyblock-link--disabled {
-    pointer-events: none;
-  }
-
   &__not-existent {
     margin: var(--space-3) 0;
   }


michele-franchi (commented):
It looks good but is there a way to disable the next arrow if the keyblock is the last one to avoid the "Requested keyblock has never been seen in the network." screen?

lukeromanowicz:
I think it should be doable by checking height against blockHeight in recentBlocksStore

janmichek:
> It looks good but is there a way to disable the next arrow if the keyblock is the last one to avoid the "Requested keyblock has never been seen in the network." screen?

Fixed and regenerated the demo

src/components/KeyblockDetailsPanel.vue
```diff
@@ -28,7 +28,32 @@
...
+                </button>
+              </app-link>
+              {{ keyblockDetails.height }}
+              <app-link :to="`/keyblocks/${keyblockDetails.height + 1}`">
+                <button
+                  :class="[
+                    'keyblock-details-panel__button',
+                    'keyblock-details-panel__button--next',
+                  ]"
+                  :disabled="!isNextKeyblockMined">
```
lukeromanowicz, line 50: the button is still clickable, even if it's disabled :/

lukeromanowicz (changes requested):
I expected that more recent keyblock button will be on the left side since every content we display is always latest first, and older stuff is available when you navigate right. But I guess this works too, just a little bit unexpected.

Overall very good job, I left only one not but I'm afraid it needs to be fixed before merging

src/components/KeyblockDetailsPanel.vue
janmichek: Good catch. You are right, haven't spotted this. Fixed

lukeromanowicz (approved):
Now it works great, thank you for the fix!

michele-franchi (approved):
It works great!

Could you please open a followup to improve arrows handling and enable again the next one if a new keyblock comes from the websocket? Not super needed right now but this would improve it.

CharisSiarampalis (approved):

PR #401, opened 2023-07-25
fix: Hash symbol inconsistency 

### LINKED ISSUE #224: Hash symbol inconsistency
### Context information

Sems like there is an unfinished idea after inheriting previous design.

So we have this `hash symbol`
image
It's present in desktop view but not present in mobile view. Which I think it's first bug

Secondly, I think the original idea behind this was somehow to help users with visual orientation. 
Eg: Transactions hashes will have this red `th` prepended, contracts will have green `ct` prepended, ... somehow color code different hashes

So we should finish the idea or drop it. Leaving this for consideration @danilosierrac 

### Steps to reproduce
all over app

### What is expected?
consistent usage of that idea

inconsistent usage

janmichek:
## Description
fixes #224 

## Demo
image

## Checklist:

- [x] remove hash symbol


CharisSiarampalis (approved):

lukeromanowicz (approved):
Thanks for removing it! Looks much cleaner and consistent with mobile view

michele-franchi (approved):
Looks good

PR #405, opened 2023-07-26
feat: make pagination button text clickable

### LINKED ISSUE #148: Make pagination text clickable

lukeromanowicz:
## Description
Resolves #148 
the text next to the pagination button is also clickable

## Demo
It works both for clicking and navigating using tabs (for accessibility reasons)
https://github.com/aeternity/aescan/assets/46789227/fd34c364-dcf8-425c-8536-5051d5bbc138

## Checklist:

### DIFF
--- a/src/components/AppPagination.vue
+++ b/src/components/AppPagination.vue
@@ -4,14 +4,14 @@
       :class="[
         'pagination__container',
         {'pagination__container--disabled': isPrevDisabled},
-      ]">
+      ]"
+      @click="!isPrevDisabled && $emit('prev-clicked')">
       <button
         :class="[
           'pagination__button',
           'pagination__button--prev',
           {'pagination__button--disabled': isPrevDisabled}
-        ]"
-        @click="$emit('prev-clicked')">
+        ]">
         <app-icon
           :size="22"
           name="caret-left"/>
@@ -24,7 +24,8 @@
       :class="[
         'pagination__container',
         {'pagination__container--disabled': isNextDisabled},
-      ]">
+      ]"
+      @click="!isNextDisabled && $emit('next-clicked')">
       <div class="pagination__label">
         {{ nextLabel }}
       </div>
@@ -33,8 +34,7 @@
           'pagination__button',
           'pagination__button--next',
           {'pagination__button--disabled': isNextDisabled}
-        ]"
-        @click="$emit('next-clicked')">
+        ]">
         <app-icon
           :size="22"
           name="caret-right"/>
@@ -76,9 +76,11 @@ defineEmits(['prev-clicked', 'next-clicked'])
     display: flex;
     align-items: center;
     color: var(--color-midnight);
+    cursor: pointer;
 
     &--disabled {
       color: var(--color-midnight-35);
+      cursor: not-allowed;
     }
   }
 


janmichek (changes requested):
I think it's not a correct implementation. This is how it should look like according to new design

image

And it should be also changed in here 
https://github.com/aeternity/aescan/pull/395

lukeromanowicz:
@janmichek I wasn't aware that changing the design was in the scope of this task since it wasn't mentioned anywhere. Please proceed with merging #395 and I will follow with the additional updates

michele-franchi:
@lukeromanowicz I thought we discussed it but yes what @janmichek suggested is correct.

src/pages/keyblocks/[id].vue
```diff
@@ -14,7 +14,7 @@
     v-if="keyblockDetails"
     :keyblock-details="keyblockDetails"/>
 
-  <app-tabs v-if="!keyblockDetails.notExistent">
+  <app-tabs v-if="keyblockDetails && !keyblockDetails?.notExistent">
```
Comment (lukeromanowicz) [inline src/pages/keyblocks/[id].vue:17]: fix for error showing up when navigating to other keyblock details pages using the navigation buttons next to keyblock height

src/styles/elements/_table.css
```diff
@@ -9,7 +9,7 @@ th {
   font-size: 14px;
   line-height: 22px;
 
-  vertical-align: initial;
+  vertical-align: middle;
```
lukeromanowicz, line 12: the buttons are slightly higher than the line height and the row content was slightly uneven between the cells

CharisSiarampalis (approved):
Seems to work although the transition when going through the keyblocks is not really smooth. It feels like the arrows and characters are shaking.

michele-franchi (approved):
Good job

src/components/PaginationButton.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <button
+    :class="[
+      'pagination-button',
+      ...($slots.default ? [`pagination-button--${direction}`] : []),
+      {'pagination-button--disabled': disabled}
+    ]"
+    @click="!disabled && $emit('click')">
```
janmichek, line 8: I think it does not have to be sanitized twice, but it causes no harm

src/components/PaginationButton.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <button
+    :class="[
+      'pagination-button',
+      ...($slots.default ? [`pagination-button--${direction}`] : []),
```
janmichek, line 5:
```suggestion
      $slots.default ? [`pagination-button--${direction}`] : [],
```

Seems like this will work the same

src/pages/keyblocks/[id].vue
Comment (janmichek) [reply src/pages/keyblocks/[id].vue:17]: I would prefer to have this in separate computed variable as that what Vue styleguide suggests

src/styles/elements/_table.css
janmichek: This has an effect on the label in mobile table header.

src/components/PaginationButton.vue
```diff
@@ -0,0 +1,77 @@
...
+  </button>
+</template>
+
+<script setup>
+import AppIcon from '@/components/AppIcon'
+
+defineProps({
+  direction: {
+    type: String,
+    default: 'left',
```
janmichek, line 28: I am not sure if the default should be left.
1. I don't thing it's a logical conclusion
2. we are not using it implicitly in any declaration
3. you could use it in the conditional class check (not checking slots)

src/components/KeyblockDetailsPanel.vue
```diff
@@ -293,25 +279,19 @@ const isNextKeyblockMined = computed(() =>
     align-items: center;
   }
 
-  &__button {
-    width: 32px;
-    height: 32px;
+  &__next-button {
```
janmichek, line 282: It's not necessarily bad, but what do you think about writing it with alignment with the child component 
```suggestion
  &__button--next {
```

janmichek (changes requested):
Works nicely 🥇 I commented some possible code improvemnents

lukeromanowicz (commented):
Thanks for the review @janmichek, all fixed

src/components/PaginationButton.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <button
+    :class="[
+      'pagination-button',
+      $slots.default ? [`pagination-button--${direction}`] : [],
```
janmichek, line 5: I think this can be simplified even more
```suggestion
      direction ? `pagination-button--${direction}` : null,
```

janmichek (changes requested):
Coolios, thanks for the fixes.

I suggested even one more improvement.

Also here I think it's not perfectly centered
image

Please take a look.

src/pages/keyblocks/[id].vue
```diff
@@ -37,6 +37,8 @@ const { keyblockDetails } = storeToRefs(keyblockDetailsStore)
 const { fetchKeyblock } = keyblockDetailsStore
 const route = useRoute()
 
+const showTabs = computed(() => keyblockDetails.value && !keyblockDetails.value?.notExistent)
```
Comment (janmichek) [inline src/pages/keyblocks/[id].vue:40]: The variable returning boolean should be `is` prefixed. Now it looks like function name on the first sight. You may accept Michelle's edit

Comment (lukeromanowicz) [reply src/pages/keyblocks/[id].vue]: Yup, done

janmichek:
Sorry I put a misleading comment, edited.

janmichek (approved):
All good now. Thank you 👍

PR #408, opened 2023-07-26
fix: remove pagination from dashboard microblock transactions panel

### LINKED ISSUE #248: When using a mobile device when selecting a microblock with one transaction it looks like it has three instead.
### Context information

application version: pr-244
device: Samsung Galaxy s22
browser: All browsers
operating system: Android

### Steps to reproduce
1.Go to the dashboard
2.Scroll down to the microblocks
3.Select on microblock with 4 transactions, then 3, 2 and finally one
4.Observe the amount of transactions that can be scrolled bellow them
### What is expected?
The amount of the transactions appearing bellow should much the number written on the box
When selecting o vox with three transactions or more only three dots can be seen. 
Then when we select a box with two the dots and scrollable transactions decrease to two.
But then when we select a box with one three dots appears again. The same transaction is repeated three times.

https://github.com/aeternity/aescan/assets/42606212/38afdfcb-e8d6-4ee4-bace-11788d93694e

### LINKED ISSUE #354: avoid screen jumping for tx-pagination in microblock section on dashboard
when clicking on the pagination in tx-block in microblock section, the screen jumps unexpectetly

don't jump around 😅 

nope

### Additional context
n/a

lukeromanowicz:
## Description
Resolves #248 
Resolves #354

Removes pagination from dashboard microblock transactions panel

## Demo

https://github.com/aeternity/aescan/assets/46789227/462541c4-ea31-48de-ad81-439c21d37613

## Checklist:


src/components/DashboardMicroblockTransactionsTable.vue
```diff
@@ -1,74 +1,49 @@
...
-const props = defineProps({
+defineProps({
   transactions: {
     required: true,
     type: Object,
   },
 })
-
-function loadPrevTransactions() {
-  fetchSelectedMicroblockTransactions(props.transactions.prev)
```
janmichek, line 60: I guess the param is not needed anymore in the function definition

janmichek (changes requested):
Good job, I like this simplification. 
One more suggested adjustment.

CharisSiarampalis (approved):

src/components/DashboardMicroblockTransactionsTable.vue
janmichek: query params does not seems to be used anywhere in the app
image

lukeromanowicz: Fixed in the latest commit

janmichek (approved):
Perfecto 👌

michele-franchi (commented):
In the desktop view pagination could be useful, why did you decide to also remove it? I don't see the same issue mobile has.

michele-franchi (approved):
I don't have a strong opinion about this solution but I also think it could work.

PR #410, opened 2023-07-26
feat: Mobile menu should support back button to close it

### LINKED ISSUE #337: Mobile menu should support back button to close it
Mobile menu should support back button to close it.

janmichek:
## Description
fixes #337 

## Checklist:

- [ ] bind navigation when the mobile menu is open


src/components/TheHeader.vue
```diff
@@ -81,6 +85,28 @@ function toggleNavigation() {
 function closeNavigation() {
   isNavigationOpen.value = false
 }
+
+function handleNavigateBack() {
+  if (isNavigationOpen.value) {
+    // simulate navigation to the same page when navigating back in browser
```
janmichek, line 91: I added several comments as the solution is a bit complex yet the simplest and clean one.
A simpler solution would be to use hash in the URL, however, we are already using it.
Although multiple hashes are allowed in URL, it's not very clean and may produce unwanted effects

michele-franchi (commented):
It seems to be working only when you are in some pages. If for example you navigate to a transaction page or the privacy policy page, you open the menu and press the back button the menu closes but you are redirected to the previous page.

https://github.com/aeternity/aescan/assets/11598532/74745f2a-c04b-4b81-ac12-a1e7b4f8e831

lukeromanowicz (changes requested):
I'm afraid it doesn't work as expected. When you take a look at the demo video after pressing back with the menu open it sets the url in browser bar to localhost:8080/ instead of the url of the previous page

janmichek:
I am afraid that the only way than is to print a #hash to the URL. But it can be messy with some other things that is using hash, eg tabs.

lukeromanowicz:
Which tabs are you refering to? If I remember correctly tabs are using get parameters, not hash

janmichek:
You are right, I will take a look agein

Issue #201, opened 2023-05-04
Move more content of account details to SSR


lukeromanowicz:
### Please describe the problem that your request should resolve.
Currently, the account details page is almost exclusively rendered on the client side. It would be nice to move viewport-independent content into SSR to improve the performance. 

### Describe the solution you'd like
viewport-independent content is rendered on the server side

PR #413, opened 2023-07-31
feat: Loading indicator for panels

### LINKED ISSUE #391: Loading indicator for panels
Add loading indicator to panels, excluding dashboard.

janmichek:
## Description
fixes #391 

## Demo

https://github.com/aeternity/aescan/assets/15363559/178b85e2-1746-48cf-8933-0141c904fac5

## Checklist:

- [x] add panel loader


janmichek:
@lukeromanowicz Can you help me to consult this please?
I tried several approaches, this works quite good
But is this loading only valid for client-side rendering - in this case only account detail?
THe other pages arrive hydrated right away.
(Don't evaluate the code or visual now)

lukeromanowicz:
@janmichek that is correct and expected. The loading indicator should not be needed for panels that are rendered on the SSR. Only those that are resolved (or reloaded with different data through navigation) on the client side need loading indicators.

Why account details panel is different, you may ask. The reason is technical debt, and there is already a task to handle this #201

lukeromanowicz:
Btw. I like the approach you took very much. Well done, mate :+1:

lukeromanowicz (changes requested):
So far it's a really good job but I think it should be applied on all panels, shouldn't it? Meaning that when you are navigating from one page to another, instead of staying on the original page you get redirected to the next page with a loading indicator displayed. At least that was my understanding of the task.

Apart from that I found one small issue with how it works: if I go to a non-existing account it will be displayed forever instead of showing the message that the account hasn't been seen in the network yet

image

michele-franchi:
Agree with @lukeromanowicz, I will wait for the review once it's fixed.

janmichek:
> So far it's a really good job but I think it should be applied on all panels, shouldn't it? Meaning that when you are navigating from one page to another, instead of staying on the original page you get redirected to the next page with a loading indicator displayed. At least that was my understanding of the task.
> 
> Apart from that I found one small issue with how it works: if I go to a non-existing account it will be displayed forever instead of showing the message that the account hasn't been seen in the network yet
> 
> image

Fixed. It was apparently some regression of MDW, the bug was present on other stages as well


https://github.com/aeternity/aescan/assets/15363559/47588baa-50e9-4791-a0e2-d60e219e3546

janmichek:
@lukeromanowicz I think I am more confused here than before.
The only entity panel where I can put loader is Account details. The other ones are SSR, so there is no moment I can put a loader since I received a hydrated page with data as we discussed before.

lukeromanowicz (commented):
As posted above - comment for ticking out review request

janmichek:
@lukeromanowicz I think I solved it by using nuxt hooks.
See the updated video demo

michele-franchi (commented):
I see the loader is applied to the whole page. My personal expectation about "loading indicator for panels" is to display the loader in the panel and keep the page structure, something like the following:

image

Can this be achieved? In the current state I see pages flashing a lot and wouldn't this interfere with tables loaders?
Maybe my expectations are too high 😄

janmichek:
> In the current state I see pages flashing a lot and wouldn't this interfere with tables loaders?

I tried to achieve this before. The problem is that the page name is included in the child id page after the loading condition.
When I try to move the loading hook to the child id page, the loader does not appear at all. 
Also, I think we are heading different direction from the initial intention of this issue (loader for panels).
If anyone has some other ideas you can take a look at the issue. I tried several approaches and ran out of ideas, but that does not mean there isn't any other way.

janmichek:
> @janmichek from my point of view the original issue is exactly about loader for panels but here I see more loader for pages instead. If I'm not wrong you achieved this with tables, the page structure remains the same and the loader is displayed inside the table.
> 
> I understand the difficulties and not sure if this is achievable. Any opinion @lukeromanowicz?

That was my previous solution. Should I revert it?

lukeromanowicz:
If I understand correctly we are discussing if the loading indicator of the panel is better or worse than showing a loading indicator of the whole page. I'd say panel indicator is better because it's less blinking, which I believe was the previous solution - right?

janmichek:
> If I understand correctly we are discussing if the loading indicator of the panel is better or worse than showing a loading indicator of the whole page. I'd say panel indicator is better because it's less blinking, which I believe was the previous solution - right?

Right. In the previous case was only implemented for account details as a client side rendered page. THe SSR-rendered pages come hydrated, so there is no moment to place the loader as it comes hydrated

lukeromanowicz (commented):
Then I'd do first panel only indicator accordingly with what we discussed during the daily

Issue #95, opened 2023-04-14
Network unavailability handling


michele-franchi:
Let user know something went wrong and provide helpful action

Now it just looks 'broken'
- unstable internet connection
- node is down
- coingecko is down
- mdw is down

anything else?

Suggested by @janmichek.

Example of a problem we want to solve here: https://github.com/aeternity/aescan/issues/127

janmichek:
The `data failed to fetch` component which was implemented just for Oracle Events had to be removed, because was colliding with loaders.
There needs to be more adjustments to distinguish to work with loaders. Ideally for the whole app, not just one entity.
https://github.com/aeternity/aescan/pull/419/commits/4487bf6df0e3819f03b8f11f82097b4a97ba3acd

janmichek:
The page should not stop working if the count is not working.
As happened in here

 https://github.com/aeternity/aescan/pull/815

PR #419, opened 2023-08-02
feat: Loading for tables

### LINKED ISSUE #150: Loading indicator for tables
I have just received feedback from @peronczyk that we could really use some loading indicator for tables

Smooth switching pages of paginated content from one page to another with some loading indication

No

### Additional context

Dashboard is excluded.

janmichek:
## Description
fixes #150 

Loader style will be improved once delivered from Danilo

## Demo

https://github.com/aeternity/aescan/assets/15363559/85cc9128-47d1-4cbe-a3f8-e0aecb6fe14f

## Checklist:

- [x] add loader to tables


lukeromanowicz:
If you don't mind I'll review it once panel loading indicators are done since there is far less changes

janmichek:
> If you don't mind I'll review it once panel loading indicators are done since there is far less changes

I don't think this dependency is needed. Can be reviewed independently

src/components/PaginatedContent.vue
```diff
@@ -109,26 +113,53 @@ const nextLabel = computed(() => {
...
 
 const handleNextClicked = () => {
+  setFixedContainerHeight()
   pageIndex.value++
   emit('next-clicked')
-  scrollToPaginatedContent()
 }
 
-const paginatedContent = ref()
+watch(
```
lukeromanowicz, line 127: I think we should make a dedicated composable for that because this will be reused in most of the pages containing tables

lukeromanowicz (commented):
This is working very nicely and apart from a minor refactoring proposition I'd accept it. My question is wether this is a concept showcase that will be applied to the remaining tables such as:

https://github.com/aeternity/aescan/assets/46789227/9b8917f5-60ae-4e74-aa94-5c1fba530f6d

or are you finished? In the second case I would add it in the other tables too :)

janmichek:
> This is working very nicely and apart from a minor refactoring proposition I'd accept it. My question is wether this is a concept showcase that will be applied to the remaining tables such as:
> 2023-08-08.09-54-28.mp4
> 
> or are you finished? In the second case I would add it in the other tables too :)

You are right, I forgot to reset the store before navigating. Fixed

src/components/PaginatedContent.vue
janmichek: I think now when I fixed the resetting it's not an issue anymore.
The pagination for tables is only handled by PaginatedContent component, right?
If I missed your point, please give me some examples of which other table would you reuse it?

michele-franchi (changes requested):
I found some issues during testing, could you please have a look at them?

- Table content changes width when switching tabs
- 500 error on page reload

https://github.com/aeternity/aescan/assets/11598532/3355059c-33c5-4d42-8d78-1b25b785906d

- Oracles events does not display the loader
Screenshot 2023-08-09 at 00 05 30

I will review again once fixed.

janmichek:
> I found some issues during testing, could you please have a look at them?
> 
>     * Table content changes width when switching tabs
> 
>     * 500 error on page reload
> 
> 
> Screen.Recording.2023-08-08.at.23.57.02.mov
> 
>     * Oracles events does not display the loader
> 
> 
> Screenshot 2023-08-09 at 00 05 30
> 
> I will review again once fixed.

Thanks for the detailed check. All fixed. It produced one more conflict, I would better resolve in here 
https://github.com/aeternity/aescan/issues/95

michele-franchi (approved):
From a UI / UX perspective it looks good. Maybe it can be improved further but that's a good base. Good job!

src/components/PaginatedContent.vue
lukeromanowicz: I misinterpreted the code. Please ignore it

lukeromanowicz (approved):
Looks great to me :+1: well done

Liubov-crypto:
LGTM

PR #421, opened 2023-08-02
feat: Add 500 error page

### LINKED ISSUE #87: Handle 500 errors
500 errors should be handled to not display the default 500 error Nuxt page, for example:

https://aescan.dev.aepps.com/transactions/th_QRAjvukXobteRXCFzZZtA23iNJTiqpX5LxZN4nhdD6vUPG96X

We need a 500 Internal Server Error page, skip illustration for now.

janmichek:
## Description
fixes #87 

## Demo
image
image

## Checklist:

- [x] Add error page

### DIFF
--- a/src/error.vue
+++ b/src/error.vue
@@ -95,7 +95,7 @@
       </app-panel>
     </div>
   </div>
-  <the-footer class="app__footer"/>
+  <the-footer/>
 </template>
 
 <script setup>
--- a/src/error.vue
+++ b/src/error.vue
@@ -70,9 +70,9 @@
       </page-header>
 
       <app-panel class="error__panel">
-        <h1 class="error__heading">
+        <h2 class="error__heading">
           Sorry! We encountered an unexpected error
-        </h1>
+        </h2>
         <p class="error__paragraph">
           ERROR CODE: 500 - Internal Server error
         </p>
@@ -106,7 +106,7 @@ import AppButton from '@/components/AppButton'
 import AppPanel from '@/components/AppPanel'
 import TheHeader from '@/components/TheHeader'
 import TheFooter from '@/components/TheFooter'
-import { APP_CREATOR, APP_DESCRIPTION, APP_KEYWORDS, APP_TITLE, APP_URL } from '~/utils/constants'
+import { APP_CREATOR, APP_DESCRIPTION, APP_KEYWORDS, APP_TITLE, APP_URL } from '@/utils/constants'
 
 useHead({
   meta: [
--- a/src/error.vue
+++ b/src/error.vue
@@ -1,8 +1,6 @@
 <template>
   <Head>
-    <Title>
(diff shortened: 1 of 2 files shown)


janmichek:
Few comments on this:

1) Hopefully, we don't have any issues causing 500 errors on production .)
The way I was developing it was to set `dev: false` on nuxt config. Then I removed some imports to cause an error.

2) the visual emoticons have not been implemented as discussed.

3) By the visual material there should be the title `Not found` but I don't think it's correct. There should be an `Internal server error` or `Unexpected Error`

4) The previous error pages have left alignment and here was the center. So I propose to keep it consistent as shown in the demo.

src/error.vue
```diff
@@ -0,0 +1,181 @@
...
+      property="og:url"
+      :content="APP_URL"/>
+    <Meta
+      name="twitter:card"
+      content="summary_large_image"/>
+    <Meta
+      name="twitter:site"
+      :content="APP_CREATOR"/>
+    <Meta
+      name="twitter:creator"
```
janmichek, line 63: these tags has not been reused from app.vue as well as the layout

lukeromanowicz (approved):
Looking really nice! well done =)

src/error.vue
```diff
@@ -0,0 +1,179 @@
...
+
+  <the-header/>
+  <div class="error">
+    <div class="error__parallax">
+      <page-header>
+        Unexpected Error
+      </page-header>
+
+      <app-panel class="error__panel">
+        <h1 class="error__heading">
```
michele-franchi, line 73: Now that I see it live I would change it to h2, to me it looks too big and also does not have a hierarchy.
If you agree let's also add a comment to the design.

src/error.vue
```diff
@@ -0,0 +1,179 @@
...
+
+<script setup>
+import { useHead } from '@vueuse/head'
+import AppLink from '@/components/AppLink'
+import PageHeader from '@/components/PageHeader'
+import AppButton from '@/components/AppButton'
+import AppPanel from '@/components/AppPanel'
+import TheHeader from '@/components/TheHeader'
+import TheFooter from '@/components/TheFooter'
+import { APP_CREATOR, APP_DESCRIPTION, APP_KEYWORDS, APP_TITLE, APP_URL } from '~/utils/constants'
```
michele-franchi, line 109: Different import? Can you unify it?

michele-franchi (changes requested):
It looks good and it's according to the design but I would slightly adjust it.

src/error.vue
janmichek: fixed

src/error.vue
janmichek: fixed

src/error.vue
janmichek: image

michele-franchi (approved):
Good job

PR #428, opened 2023-08-06
feat: account activities

### LINKED ISSUE #10: Account activities tab
Columns:
- Hash: in some cases we won’t have any tx-hash, e.g. AENS expiration won’t have any tx-hash, the same for Beneficiary reward
- Date
- Type
- Activity
- Data

Type | Activity | Data | Technical considerations
-- | -- | -- | --
AE | Transferred toReceived fromBeneficiary reward | ak_… + Æ amountak_… + Æ amountÆ amount | how to deal with SH-DEX activities to avoid duplications?
AENS | ClaimedExpiredRevokedReceived fromTransferred to | xyz.chainxyz.chainxyz.chainak_… + xyz.chainak_… + xyz.chain |  
AEX-9 | Token createdTransferred toReceived fromMint receivedBurned | TOKEN_SYMBOLak_… + TOKEN_SYMBOL & amountak_… + TOKEN_SYMBOL & amountTOKEN_SYMBOL & amountTOKEN_SYMBOL & amount | how to deal with SH-DEX activities to avoid duplications?can we distinguish transfers, mints and burns?
AEX-141 | Collection createdTransferred toReceived fromMint receivedBurned | COLLECTION_NAMEak_… + COLLECTION_NAME & nft_idak_… + COLLECTION_NAME & nft_idCOLLECTION_NAME & nft_idCOLLECTION_NAME & nft_id | can we distinguish transfers, mints and burns?
SH-DEX(to be checked) | SwappedAdded liquidityRemoved liquidity | TOKEN_SYMBOL & amount ↔︎ TOKEN_SYMBOL & amountTOKEN_SYMBOL & amount +TOKEN_SYMBOL & amount ↔︎ POOL_SYMBOL & amountPOOL_SYMBOL ↔︎ TOKEN_SYMBOL & amount + TOKEN_SYMBOL & amount | see AE and AEX-9 comments, avoid duplicated outputmaybe we need to include that in the type “AEX-9” and cannot have it separately in SH-DEX as type
? | ? | ? | ?

<p></p>

Endpoint: GET /v2/accounts/:id/activities

lukeromanowicz:
## Description
Resolves #10 
Implements activities tab.

Doesn't implement:
* detailed AEX-9 and AEX-141 information for contract create and call transactions for these Tokens - MDW tickets https://github.com/aeternity/ae_mdw/issues/1497, https://github.com/aeternity/ae_mdw/issues/1505, and https://github.com/aeternity/ae_mdw/issues/1523
* activity duplication removal mechanism - hiding internal transfers on client side related to the transactions would break the pagination. - MDW tickets  https://github.com/aeternity/ae_mdw/issues/1504 and
* fix for https://github.com/aeternity/ae_mdw/issues/1498
* activities count (aka pagination information at the top of the table) - MDW ticket https://github.com/aeternity/ae_mdw/issues/1506

I haven't created a followup tickets for the aescan yet because I would like us first to discuss on how to split the work

## Demo

Uploading firefox_wm9bqQCzZa.mp4…

## Checklist:


michele-franchi (commented):
I had a quick look at the demo but It looks like it's not according do the specifications:
https://github.com/aeternity/aescan/issues/10#issuecomment-1653418775

In particular related to "Type" and "Activity" columns.

lukeromanowicz:
@michele-franchi I fixed the TYPE column for most of the activities apart from token contract calls which are indistinguishable from custom smart contracts as stated in the description.

src/components/AccountActivitiesPanel.vue
```diff
@@ -0,0 +1,73 @@
...
+import PaginatedContent from '@/components/PaginatedContent'
+
+const accountStore = useAccountStore()
+const { fetchAccountActivities } = accountStore
+const { accountActivities, accountDetails } = storeToRefs(accountStore)
+
+const limit = computed(() => isDesktop() ? 10 : 3)
+
+function loadPrevActivities() {
+  fetchAccountActivities({
```
janmichek, line 38: Here I wanted to ask why we omit signalizing the async nature of functions? My IDE is warning me that it's missing there. I think it's generally good idea to explicitly use await / async to let user (programmer) know about the nature of the data even tho it is working without it. In my eyes it's more confusing than beneficial.

src/components/AccountActivitiesTable.vue
```diff
@@ -0,0 +1,105 @@
...
+          <hint-tooltip>
+            {{ accountHints.activityData }}
+          </hint-tooltip>
+        </th>
+      </tr>
+    </thead>
+    <tbody>
+      <tr
+        v-for="(activity, index) in accountActivities?.data"
+        :key="`${activity.hash}-${activity.type}-${index}`">
```
janmichek, line 40:
```suggestion
        :key="index">
```
Index on its own should do the job

src/components/InternalTransferCell.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <span v-if="activity.payload?.kind !== 'reward_block'">N/A</span>
+  <template v-else>
+    <app-chip size="sm">
+      {{ formatAePrice(formatAettosToAe(activity.payload.amount)) }}
+    </app-chip>
+  </template>
```
janmichek, line 7:
```suggestion
 <template v-if="activity.payload?.kind !== 'reward_block'">
    N/A
  </template>
  <app-chip
    v-else
    size="sm">
    {{ formatAePrice(formatAettosToAe(activity.payload.amount)) }}
  </app-chip>
```

I think it can be achieved with less code

src/stores/accountDetails.js
```diff
@@ -90,6 +96,12 @@ export const useAccountStore = defineStore('account', () => {
     accountNamesCount.value = data.data.length
   }
 
+  async function fetchAccountActivities({ accountId, limit, queryParameters } = {}) {
+    const defaultParameters = `/v2/accounts/${accountId}/activities?limit=${limit ?? 10}`
```
janmichek, line 100: For the loadings, here would be good to reset data to null

src/utils/hints/accountHints.js
```diff
@@ -13,4 +13,9 @@ export const accountHints = {
   name: 'Name owned by this account.',
   expires: 'Keyblock height and estimated date and time when the name owned by the account expires.',
   pointsTo: 'The account that points to the name owned by the account.',
+  activityHash: 'Transaction hash of the transaction corresponding to the activity where the account was involved. Some of the activities are executed internally and therefore don\'t have a transaction hash.',
+  activityTime: 'Keyblock height and exact date and time when the activity has been executed by inclusion into a microblock.',
+  activityType: 'Type of the activity where the account was involved.',
+  activityMeaning: 'The meaning of the activity.',
+  activityData: 'The data displayed differs for each activity type.',
```
janmichek, line 20: here it's not easy for me to understand after reading it. Could you try something that would explain it for dummies? .)

src/utils/hints/accountHints.js
```diff
@@ -13,4 +13,9 @@ export const accountHints = {
   name: 'Name owned by this account.',
   expires: 'Keyblock height and estimated date and time when the name owned by the account expires.',
   pointsTo: 'The account that points to the name owned by the account.',
+  activityHash: 'Transaction hash of the transaction corresponding to the activity where the account was involved. Some of the activities are executed internally and therefore don\'t have a transaction hash.',
+  activityTime: 'Keyblock height and exact date and time when the activity has been executed by inclusion into a microblock.',
```
janmichek, line 17: here it's not easy for me to understand after reading it. Could you try something that would explain it for dummies? .)

janmichek (changes requested):
Works superb 🚗  given the current limitations.
I added some code suggestions

src/components/AccountActivitiesPanel.vue
lukeromanowicz: @janmichek Since this function is executed only on the client side and upon execution nothing is awaiting it, there wasn't any difference between returning fetchAccountActivities and just executing it. That being said of course it's a good practice to do so. I'll fix it

src/components/AccountActivitiesTable.vue
lukeromanowicz: I did it this way because it's not a good practice to use the index alone https://stackoverflow.com/questions/44531510/why-not-always-use-the-index-as-the-key-in-a-vue-js-for-loop

src/components/InternalTransferCell.vue
lukeromanowicz: Oh, right, of course!

src/utils/hints/accountHints.js
lukeromanowicz: I literally copied it from the account transaction's time hint above :D I would keep it this way for consistency or improve it everywhere (but I don't have a good idea how)

src/utils/hints/accountHints.js
lukeromanowicz: Same as above. It's copied from the hint above.

src/components/AccountActivitiesTable.vue
janmichek: Aha, thanks. I was not aware of that. So the nature of data is live at and any moment new record can pop up in the table and break the keys, right?

src/utils/hints/accountHints.js
janmichek: Ok, as you wish. For me was not easy to understand

src/components/AccountActivitiesTable.vue
lukeromanowicz: Well... it's not live in the sense that it's fetched through WebSockets but loading other pages will put another set of records into the array with the same indexes but different content

janmichek: Would this be solved by resetting store data before fetching new ones as I pointed out below?

lukeromanowicz:
@janmichek thanks for the review. I think I addressed everything.

src/components/AccountActivitiesTable.vue
lukeromanowicz: I used that long key out of the habit of not setting an index as the key. In this specific example, index for the key is just fine for the reason you mentioned so I simplified it.

https://vuejs.org/api/built-in-special-attributes.html#key

janmichek (approved):
Crazy cool 😎 Thanks for the collaboration

janmichek:
@lukeromanowicz Question
So there is the task to eliminate duplicates https://github.com/aeternity/ae_mdw/issues/1498 but on the other hand you said they make sense and you wanted me to identify them by Dex contract ID

lukeromanowicz:
@janmichek although ae_mdw#1498 has "duplicates" in the title, it's unrelated to identifying contract calls as SH-DEX or not. It was a bug when you send a transfer to yourself, you would see that event twice. I guess you can ignore that task completely since it doesn't impact meaningfully the activities task because no one sends transactions to themselves :D

janmichek:
The demo video has been regenerated
Some accounts to test

ak_dArxCkAsk1mZB1L9CX3cdz1GDN4hN84L3Q8dMLHN4v8cU85TF
ak_2r2xGN1v8QKQDNk35WBKyH6Wxz1WP3pxhBNrLgeaK9aTybjvDL
ak_gZ55FLJoGEfF5gJ2xJU7j9uSH3fEjoTZq3ygQG8gYA4no6GFW
ak_2epACQAj2w8EBX4AVodp3c4NE5N4HVMRLW7KvsA2ncbRV4QNbT

michele-franchi:
@marc0olo it would be nice to also have your review

src/components/AccountActivityCell.vue
```diff
@@ -0,0 +1,104 @@
...
+  case 'InternalTransferEvent':
+    if (props.activity.payload.kind === 'reward_block') {
+      return 'Beneficiary reward'
+    }
+    if (SH_CONTRACTS.includes(props.activity.payload.contractId)) {
+      return 'SH-DEX transfer'
+    }
+    return 'Internal transfer'
+  case 'Aex9TransferEvent':
+    return 'Trasnfered to'
```
marc0olo, line 97:
```suggestion
    return 'Transferred to'
```

marc0olo (changes requested):
## General

- Are you sure that this is aligned with what's shown in the screenshot here?
   - https://github.com/aeternity/aescan/issues/10#issuecomment-1653418775
- It would be great to differentiate the activity types somehow with a color to make immediately clear what activity it is
- I think it needs to be clarified what the purpose of this tab is. Currently it is very technical, whereas I thought it is very user-centric and easily understandable for everybody.

If we really want to show everything it needs to be better understandable by the user IMO. @michele-franchi 

## Beneficiary Reward
image

Why does the data column include `reward_block`? The AE amount should be sufficient.

## AE Transfers
image
- not sure, but it seems like we are just using the data in the regular "tx"-table whereas we have an account context here.

## DEX

### Swap
image
- not aligned with the original requirements. data not really meaningful. if possible we should see the tokens that have been swapped

## AENS
image
- do we really want to show everything here? 🤔 (refund etc.)
   - from my personal point of view this isn't an "activity" that I have performed
   - I would even hide "Preclaim" here 🤔 
- the name claim is missleading
    - in the original requirements I seem to have missed the "Auction Bid" activity which is basically a claim in an active auction.
- I would prefer showing meaningful data here that the user can actually easily understand. otherwise the activity tab confuses me personally a lot

lukeromanowicz:
@marc0olo 

> Are you sure that this is aligned with what's shown in the screenshot here?
Some of the details cannot be aligned because `*ContractCall*Events` usually do not have metadata about the contract details   
https://github.com/aeternity/ae_mdw/issues/1497

This is why we don't remove "internal events" that were initially considered duplicates, as they have vital information. Removing them is not only not possible for technical reasons (filtering them out breaks the pagination) but also because there are some cases where they appear in the activities even though they were triggered by a smart contract call by another account.

> I think it needs to be clarified what the purpose of this tab is. Currently it is very technical, whereas I thought it is very user-centric and easily understandable for everybody.

Let's dive deeper into what parts you would suggest to simplify and what is the outcome you'd expect.

> not aligned with the original requirements. data not really meaningful. if possible we should see the tokens that have been swapped
As I wrote above, unfortunately, there isn't sufficient information to display anything more in the smart contract call.  The details you are mentioning are in the activities that follow smart contract call. There can be a lot of activities caused by one smart contract call, but there can be also 0 activities even a smart contract call was performed. It all depends on the smart contract call details and its result which is unknown for the smart contract call event itself.

janmichek:
@marc0olo 
Thanks for the great review. Most of it has been changed. I provide some more commentary below


> It would be great to differentiate the activity types somehow with a color to make immediately clear what activity it is

Not sure if this would bring more good than harm, but we can discuss it


> Why does the data column include reward_block? The AE amount should be sufficient.

removed



> not sure, but it seems like we are just using the data in the regular "tx"-table whereas we have an account context here.

made it more readable



> not aligned with the original requirements. data not really meaningful. if possible we should see the tokens that have been swapped

As Luke pointed we are waiting for middleware in this. Created followup https://github.com/aeternity/aescan/issues/512

nuxt.config.ts
```diff
@@ -36,6 +36,7 @@ export default defineNuxtConfig({
       ALTERNATIVE_NETWORK_NAME: process.env.ALTERNATIVE_NETWORK_NAME,
       AE_TOKEN_ID: process.env.AE_TOKEN_ID,
       DEBUG_MODE: process.env.DEBUG_MODE,
+      SH_CONTRACTS: process.env.SH_CONTRACTS?.split(';'),
```
michele-franchi, line 39: I would call this `SH_DEX_CONTRACTS` to make it more explicit.

michele-franchi (approved):
I think this is already a good improvement and I would merge this and further improve separately because the feature is big and there are a lot of different activities.

The only visual thing that could be improved is that names are truncated even when there is enough space, for example:
image

I would suggest to open a followup for this with @marc0olo suggestions that were not implemented.

nuxt.config.ts
janmichek: Okay, good. I renamed that and asked Petar to rename the variable for deployment environments. So maybe for some short time period, it will not be working. I will inform you about the current status

janmichek:
> I think this is already a good improvement and I would merge this and further improve separately because the feature is big and there are a lot of different activities.
> 
> The only visual thing that could be improved is that names are truncated even when there is enough space, for example: image
> 
> I would suggest to open a followup for this with @marc0olo suggestions that were not implemented.

Truncation of values removed

janmichek:
> I think this is already a good improvement and I would merge this and further improve separately because the feature is big and there are a lot of different activities.
> 
> The only visual thing that could be improved is that names are truncated even when there is enough space, for example: image
> 
> I would suggest to open a followup for this with @marc0olo suggestions that were not implemented.

Mostly has been fixed, the only followup is this one https://github.com/aeternity/aescan/issues/512

nuxt.config.ts
janmichek: Seems to be good now

Liubov-crypto (commented):
I have a question: is this normal view for this type of Tx? 
scan

janmichek:
@Liubov-crypto 

Thanks for the review with screenshots.

> 1. is this normal view for this type of Tx?

Yes this is expected result. You will find the explanation here https://github.com/aeternity/ae_mdw/issues/1504#issue-1838201581


> 2. Also if you check this account: ak_1BFuEjDzz4Qz5JK85H7Q4dRjMB6yKTf28BA9oCZjZGjDdT4aF it contains different activities and it's strange to see the amount mixed with the text. Is this acceptable?

Yes, the mixed content is intentional as the column is named `Data`. Within the context of row makes sense.


> 3. Idk if it's related to this pr, but is this view ok? (https://pr-428-aescan.stg.aepps.com/contracts/ct_KeTvHnhU85vuuQMMZocaiYkPL9tkoavDRT3Jsy47LK2YqLHYb)


I am not sure what you referring to here, but on the first sight looks as expected



I see you are using words like 'ok' and 'normal'. To me it seems to broad and subjective description. Some more descriptive wording would be appreciated, so there is no space for misunderstanding. Thank you

Liubov-crypto:
@janmichek thank you for the explanation! In this case, LGTM.

janmichek:
Let's merge this as it brings a lot of good value to users. It's a big issue so let's not delay it that much. We are willing to improve this in the future @marc0olo

PR #434, opened 2023-08-08
feat: lexon token listing

### LINKED ISSUE #427: add LAEX (Lexon.org) token to the default tokens
I want to see the `LAEX` token (contractId: `ct_xtk8rSz9suPb6D6VLquyfVji25FcnFRDjn3dnn5mmvHsPiESt`) added to the default token list shown at https://aescan.io/tokens

### Additional context
- the token contract is currently not classified by the mdw, but this should be fixed early next week afaik (cc @jyeshe @sborrazas)

Please use following icon for the token:
lexon-logo-black

michele-franchi:
## Description

Fixes #427.

As we already know this is a temporary solution until we handle tokens listing properly.

## Demo

image

## Checklist:


lukeromanowicz (approved):
Looks good, and works well too. I hope DEX will be updated soon to include it because for a long term, it's going to be hell in terms of maintenance

marc0olo:
> I hope DEX will be updated soon to include it because for a long term, it's going to be hell in terms of maintenance

for long term even dex is not the right place IMO. we need sth. similar to https://tokenlists.org that every client can consume independently

janmichek (approved):
Ok, cool. I added one more commit. I noticed an unstyled class which was residue from my previous work.

Setting up follow-up issue would be nice so we don't forget.

PR #441, opened 2023-08-09
feat: Loading indicator for panels  2

### LINKED ISSUE #391: Loading indicator for panels

janmichek:
## Description
fixes #391 

## Demo

https://github.com/aeternity/aescan/assets/15363559/ebaaa15e-aa13-4914-944c-35756209e1b3

## Checklist:

- [x] add panel loader

### DIFF
--- a/src/composables/useLoading.js
+++ b/src/composables/useLoading.js
@@ -8,5 +8,5 @@ export default function useLoading() {
   nuxtApp.hook('page:finish', () => {
     isLoading.value = false
   })
-  return isLoading
+  return { isLoading }
 }
--- a/src/composables/useLoading.js
+++ b/src/composables/useLoading.js
@@ -1,0 +1,12 @@
+export default function useLoading() {
+  const nuxtApp = useNuxtApp()
+  const isLoading = ref(true)
+
+  nuxtApp.hook('page:start', () => {
+    isLoading.value = true
+  })
+  nuxtApp.hook('page:finish', () => {
+    isLoading.value = false
+  })
+  return isLoading
+}
--- a/src/pages/state-channels/index.vue
+++ b/src/pages/state-channels/index.vue
@@ -16,9 +16,20 @@
     </template>
   </page-header>
 
-  <state-channels-panel/>
+  <state-channels-panel v-if="!isLoading"/>
+  <loader-panel v-else/>
 </template>
 
 <script setup>
 import { stateChannelsHints } from '@/utils/hints/stateChannelsHints'
+
+const nuxtApp = useNuxtApp()
+const isLoading = ref(true)
+nuxtApp.hook('page:start', () => {
+  isLoading.value = true
+})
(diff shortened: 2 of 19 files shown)


src/pages/state-channels/index.vue
```diff
@@ -16,9 +16,20 @@
...
 
-  <state-channels-panel/>
+  <state-channels-panel v-if="!isLoading"/>
+  <loader-panel v-else/>
 </template>
 
 <script setup>
 import { stateChannelsHints } from '@/utils/hints/stateChannelsHints'
+
+const nuxtApp = useNuxtApp()
```
lukeromanowicz, line 26: I propose to create a composable in `src/composables/useLoading` that will return `isLoading` ref and then reuse it in every component.

lukeromanowicz (changes requested):
For now, it works pretty decently but it involves slight code duplication that we can easily reduce by creating a composable. In the future we could probably add some animation to make it even nicer but I don't think we should focus on that right now

src/pages/state-channels/index.vue
janmichek: Cool suggestion, thanks for the tip .) moved there

michele-franchi (approved):
It looks really good to me. The only thing I noticed is that the vertical padding in the panel looks less then the one applied with tables so I would keep the same one but I approve in advance.

janmichek:
> It looks really good to me. The only thing I noticed is that the vertical padding in the panel looks less then the one applied with tables so I would keep the same one but I approve in advance.

Good catch, adjusted

src/composables/useLoading.js
```diff
@@ -0,0 +1,12 @@
...
+  const nuxtApp = useNuxtApp()
+  const isLoading = ref(true)
+
+  nuxtApp.hook('page:start', () => {
+    isLoading.value = true
+  })
+  nuxtApp.hook('page:finish', () => {
+    isLoading.value = false
+  })
+  return isLoading
```
lukeromanowicz, line 11: Please return the ref in an object to make it usable in the same way as all other composables (e.g. from vueUse)

lukeromanowicz (approved):
I suggested one minor improvement for the sake of consistency and I think it would be also nice to somehow preload or at least cache the loading indicator svg so it's visible in the very beginning just like the loading text.

That being said these are minor stuff that we can also fix with a follow-up task if you want.


https://github.com/aeternity/aescan/assets/46789227/d3ea9344-3dc5-4ac7-b310-cae215683495

src/composables/useLoading.js
janmichek: Ok, done

lukeromanowicz:
@janmichek please ignore the part about caching... I had "disable caching" enabled :man_facepalming:

Liubov-crypto:
LGTM

PR #444, opened 2023-08-10
feat: implement not found panel for all details pages

### LINKED ISSUE #106: Handle not found pages
Users may enter an entity details page with a wrong id put into the parameter. 

For example someone could directly type an invalid contract id like this:
https://aescan.dev.aepps.com/tokens/ct_J3zBY8xxjsRr3QojETNw48Eb38fjvEuJKkQ6KzECvubvEcvAE

The solution could simply be to redirect the user to a 404 not found page but we need a design for it.

Every page should be specific to the entity.

lukeromanowicz:
## Description
Resolves #106 

## Demo
image

## Checklist:


src/components/NotFoundPanel.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <app-panel class="not-found-panel">
```
janmichek, line 2: Can this panel also be reused in pages/error/query ?

src/pages/contracts/[id].vue
```diff
@@ -54,12 +66,21 @@ const { isLoading } = useLoading()
 
 await useAsyncData(() => fetchContractDetails(route.params.id))
 
-if (process.client) {
+const contractFound = ref(true)
```
Comment (janmichek) [inline src/pages/contracts/[id].vue:69]:
```suggestion
const isContractFound = ref(true)
```

similar to other occurences

src/pages/contracts/[id].vue
```diff
@@ -15,7 +19,7 @@
       </app-link>
     </template>
   </page-header>
-  <template v-if="!isLoading">
+  <template v-if="!isLoading && contractFound">
```
Comment (janmichek) [inline src/pages/contracts/[id].vue:22]:
```suggestion
  <template v-if="hasError">
```



there should be simple expressions in templates so this should go to computed https://vuejs.org/style-guide/rules-strongly-recommended.html#simple-expressions-in-templates

src/components/NotFoundPanel.vue
```diff
@@ -0,0 +1,67 @@
...
+  &__button {
+    width: 215px;
+    padding: var(--space-3) var(--space-5);
+  }
+}
+</style>
+
+<style>
+.not-found-panel {
+  q {
```
janmichek, line 61: 1. This is not written according to BEM methodology
2. I wouldn't say it's necessarily a quote by it's meaning and by style - quotes are usually italics. I would use the same as `search-error__query`. That's ideal scenarion
3. but in the case we want to save some code so we don't need to be that dogmatic. In that case, I suggest styling it in separate stylesheet styles/elements/_q.css

src/pages/microblocks/[id].vue
```diff
@@ -5,18 +5,30 @@
...
-  <template v-if="!isLoading">
+  <template v-if="!isLoading && microblockFound">
     <microblock-details-panel
       v-if="microblockDetails"
       :microblock-details="microblockDetails"/>
 
     <microblock-transactions-panel/>
   </template>
-  <loader-panel v-else/>
+  <loader-panel v-else-if="isLoading"/>
```
Comment (janmichek) [inline src/pages/microblocks/[id].vue:24]: I personally I prefer to have more flat logic to ease of readability - means not using else-if if possible. But I am fine with this solution if you see things differently

src/pages/keyblocks/[id].vue
```diff
@@ -5,12 +5,17 @@
...
+      not found
+    </template>
+
     <template #tooltip>
       {{ keyblocksHints.keyblock }}
     </template>
   </page-header>
 
-  <template v-if="!isLoading">
+  <template v-if="!isLoading && keyblockFound">
```
Comment (janmichek) [inline src/pages/keyblocks/[id].vue:18]: In this case wrapping by template just to chain condition it's complicating the readability. It's an extension of an idea I expressed in `src/pages/microblocks/[id].vue` I think it's too much logic in template. I think the better place for logic is in script block where would be nicely explained by naming. I would suggest something like.

````
 <keyblock-details-panel
      v-if="!isDetailsLoading"
      :keyblock-details="keyblockDetails"/>
````

or


````
 <keyblock-details-panel
      v-if="isDetailsLoaded"
      :keyblock-details="keyblockDetails"/>
````

File: src/pages/state-channels/[id].vue [hunk 7]
```diff
@@ -6,6 +6,10 @@
   <page-header>
```
Comment (janmichek) [inline src/pages/state-channels/[id].vue:6]: https://github.com/aeternity/aescan/assets/15363559/a364f40a-6f4f-4a8e-9283-8915582ab480

Does not seems to work here

janmichek (changes requested):
Cool, good job! Mostly working good, apart from state channels. 
I also provided some code suggestion to improve readability.

src/pages/microblocks/[id].vue
Comment (lukeromanowicz) [reply src/pages/microblocks/[id].vue:24]: Well I could make one more nesting level but I'm not sure if it's actually better for the code readability. @michele-franchi wdyt?

src/pages/state-channels/[id].vue
Comment (lukeromanowicz) [reply src/pages/state-channels/[id].vue:6]: My bad, I must have broke it during pre-push cleanup

src/components/NotFoundPanel.vue
lukeromanowicz: I'll give it a try

src/pages/contracts/[id].vue
Comment (lukeromanowicz) [reply src/pages/contracts/[id].vue]: Thank you, fixed

src/pages/state-channels/[id].vue
Comment (lukeromanowicz) [reply src/pages/state-channels/[id].vue:6]: Thank you, fixed

src/pages/contracts/[id].vue
Comment (lukeromanowicz) [reply src/pages/contracts/[id].vue]: I don't like the result of following this rule in this specific scenario. The following example looks more readable to me:
image

Rather than scrolling up and down the file to see what logic is hidden behind the proposed computed.

src/pages/keyblocks/[id].vue
Comment (lukeromanowicz) [reply src/pages/keyblocks/[id].vue]: This comment also applies here: https://github.com/aeternity/aescan/pull/444#discussion_r1293317241

lukeromanowicz:
@janmichek please let me know if you find my comments convincing or if you still insist on refactoring into computeds.

michele-franchi (changes requested):
The functionality looks good but I see some visual differences, compared to designs.

The design is the following (for example name):
image

The title is always "Not Found", which I think it's better and the button does not take two lines:

image

src/pages/microblocks/[id].vue
Comment (michele-franchi) [reply src/pages/microblocks/[id].vue:24]: What would be the other option? To simply use v-if instead of v-else-if?
In any case let's add what we pick to the code guidelines if it isn't already there.

lukeromanowicz:
@michele-franchi 
I understood during the PBR that the error page designs are a loose suggestion of how it should look and that we should not implement any additional images or emoticons at the current stage.

Moreover, the task specified:

> Every page should be specific to the entity.

I implemented it this way mostly because I assumed the titles and their hints should indicate what was not found. Thank you for the additional explanation - I'll adjust it accordingly.

Not sure why the button's CSS in production mode looks differently. Probably the CSS loading order strike again

michele-franchi:
@lukeromanowicz 

The design should always be the source of truth 😄  The discussion about adjustments / emoticons was related to the 500 error page, not the 404 one.

Any opinion about the other question I asked?

> I'm also wondering if there isn't a way to handle this as a separate default page instead of adding logic to every single page? For example I see [here](https://nuxt.com/docs/getting-started/error-handling#example-1) errors are thrown and handled in other pages.

lukeromanowicz:
@michele-franchi Oh, I see. Sorry for getting it wrong. 

Yeah, if the title is always the same and we don't want to feed any entity-specific hints then it makes sense to unify it into a single error page.

src/pages/contracts/[id].vue
Comment (janmichek) [reply src/pages/contracts/[id].vue]: Thanks, looks better. If you want to make it even nicer, I suggest to sort the variables in [id] pages (optional just for consistency). They are in different order in almost every page

src/components/NotFoundPanel.vue
janmichek: So what is the outcome?

src/pages/microblocks/[id].vue
Comment (janmichek) [reply src/pages/microblocks/[id].vue:24]: No need to panic and rewrite all guidelines :D I started the sentence with my personal preference. However, I will come up with an example.

janmichek (approved):
Niceee, thanks for the partial fixes. Some points has not been answered + adding of example proposal of my idea.

src/components/NotFoundPanel.vue
lukeromanowicz: I think it worked :+1:

lukeromanowicz:
@lukeromanowicz @janmichek I rewrote everything to the new error-handling approach we discussed during the daily. Please check it out and let me know what you think.

src/errors/PageNotFoundError.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <Head>
+    <Title>{{ APP_TITLE_SHORT }} | Page not found</Title>
+  </Head>
+
+  <page-header>
+    Page not found
```
michele-franchi, line 7: Should be `Page Not Found` to keep consistency

src/errors/SearchNotFoundError.vue
```diff
@@ -0,0 +1,31 @@
+<template>
+  <Head>
+    <Title>{{ APP_TITLE_SHORT }} | Search not found</Title>
+  </Head>
+
+  <page-header>
+    Search not found
```
michele-franchi, line 7: Should be `Search Not Found` for consistency

src/errors/UnexpectedError.vue
```diff
@@ -71,7 +64,7 @@ const route = useRoute()
 
   &__button {
     width: 215px;
-    padding: var(--space-3) var(--space-5);
+    color: #fff !important;
```
michele-franchi, line 67: Is the important needed? Any other way to achieve it?

src/pages/tokens/[id].vue
```diff
@@ -48,8 +48,20 @@ const { fetchTokenDetails } = tokenDetailsStore
 
 const { isLoading } = useLoading()
 
-await fetchTokenDetails(route.params.id)
-
+try {
+  await fetchTokenDetails(route.params.id)
+} catch (error) {
+  if ([400, 404].includes(error.response?.status)) {
```
Comment (michele-franchi) [inline src/pages/tokens/[id].vue:54]: What happens if the status is 500? It also applies to the other details screens.

michele-franchi (changes requested):
Good job, I left some comments.

When you are already in a Not Found page and you search for something that does not exists it does not redirect you to the search not found, instead you see this error:
image

src/components/NotFoundPanel.vue
```diff
@@ -0,0 +1,64 @@
...
+    @media (--desktop) {
+      &:last-of-type {
+        margin-bottom: var(--space-6);
+      }
+    }
+  }
+
+  &__button {
+    width: 215px;
+    color: #fff !important;
```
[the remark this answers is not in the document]
lukeromanowicz: Yes, it's needed. Copied it from the previous place pretty much unchanged.

src/errors/UnexpectedError.vue
lukeromanowicz: Yes, it's needed. It was like this before and I left it unchanged. I believe there is a separate task to handle these issues #7

src/pages/tokens/[id].vue
Comment (lukeromanowicz) [reply src/pages/tokens/[id].vue:54]: Good catch, `throw error` is missing to show UnexpectedError page

src/pages/contracts/[id].vue
Comment (lukeromanowicz) [reply src/pages/contracts/[id].vue]: It has been simplified in the new version

src/error.vue
```diff
@@ -1,109 +1,98 @@
...
+        content="summary_large_image"/>
+      <Meta
+        name="twitter:site"
+        :content="APP_CREATOR"/>
+      <Meta
+        name="twitter:creator"
+        :content="APP_CREATOR"/>
+    </Head>
+
+    <NuxtErrorBoundary>
```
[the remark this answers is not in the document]
lukeromanowicz: Why this can't be simplified, you may ask. Great question. It's because on ALL of our error pages, we have still a fully interactive application with links, searches, and data from APIs which may again trigger some errors. If we simplified Error 500 to the bare minimum, all the other pages could be implemented through a single NuxtErrorBoundary on the app level, and all of that extra innerError logic would be removed.

Fortunately, the only "ugly" code is in the error.vue. Everything else looks clean to me.

janmichek (changes requested):
Thanks for the solid solution 🦾  I have some questions and suggested some improvements.

src/components/AppButton.vue
```diff
@@ -40,6 +40,11 @@ defineProps({
     type: String,
     default: '_blank',
   },
+  size: {
+    type: String,
+    default: 'md',
+    validator: value => ['sm', 'md'].includes(value),
```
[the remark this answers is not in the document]
lukeromanowicz: Please explain why this should be lg. It's counterintuitive to me that 'lg' is supposed to be the default value instead of md

src/errors/EntityNotFoundError.vue
```diff
@@ -0,0 +1,40 @@
+<template>
```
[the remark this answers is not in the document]
lukeromanowicz: Very good idea, thanks

src/pages/keyblocks/[id].vue
```diff
@@ -45,10 +46,21 @@ const { isLoading } = useLoading()
...
   return true
 })
+
+if (error.value) {
+  throw showError({
+    data: {
+      entityId: route.params.id,
+      entityName: 'Keyblock',
+    },
+    statusCode: 404,
```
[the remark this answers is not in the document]
Comment (lukeromanowicz) [reply src/pages/keyblocks/[id].vue:60]: Actually, this can even be skipped since 404 is applied anywhere by the specific error handler component. Thanks

src/components/AppButton.vue
```diff
@@ -3,7 +3,7 @@
     v-if="!!to"
     :to="to"
     :target="target"
-    class="button">
+    :class="['button',`button--size-${size}`]">
```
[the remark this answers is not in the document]
lukeromanowicz: I wasn't aware that other button sizes got removed. I'm applying the default size since it's the only one we use in the app

src/components/AppButton.vue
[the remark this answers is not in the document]
lukeromanowicz: Please ignore the above comment. I wasn't aware that other button sizes got removed. I'm applying the default size to all buttons since it's the only one we use in the app

michele-franchi (commented):
It looks good but I noticed the following issue:

If you navigate to a not found page, example: https://pr-444-aescan.stg.aepps.com/transactions/hello
you get the correct screen but if you type something in the search bar which does not exist like `testtesttest`, the url does not change. Maybe this will be solved in the upcoming enhanced search PR?

lukeromanowicz:
@michele-franchi why should the url change if it would still lead to something that doesn't exist? We don't want to have a malfunctioning URL in our browser history, do we? Otherwise, yes, I would adjust it accordingly in the search-related tickets since this search mechanism will change anyway

src/errors/EntityDetailsNotFoundError.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <Head>
+    <Title>{{ APP_TITLE_SHORT }} | {{ errorData.entityName }} not found</Title>
+  </Head>
+
+  <page-header>
+    Not Found
```
janmichek, line 7: Could be even like this?

```suggestion
    {{ errorData.entityName }} Not Found
```

src/errors/SearchNotFoundError.vue
```diff
@@ -0,0 +1,38 @@
...
+  <Head>
+    <Title>{{ APP_TITLE_SHORT }} | Search Not Found</Title>
+  </Head>
+
+  <page-header>
+    Search Not Found
+  </page-header>
+
+  <not-found-panel>
+    Oops! We are sorry. The search string
```
janmichek, line 11:
```suggestion
    Oops! We are sorry. The search query
```

src/errors/UnexpectedError.vue
```diff
@@ -1,56 +1,44 @@
...
-  <app-panel class="search-error__panel">
-    <p class="search-error__paragraph">
-      Oops! We are sorry. The search string
-      <span class="search-error__query">
-        {{ route.params.query }}</span>
-      was not found.
+  <app-panel class="unexpected-error__panel">
+    <h2 class="unexpected-error__heading">
+      Sorry! We encountered an unexpected error
+    </h2>
```
janmichek, line 9: Could you reuse `heading` slot?

src/errors/UnexpectedError.vue
```diff
@@ -1,56 +1,44 @@
...
-})
-
-const route = useRoute()
+setResponseStatus(500, 'Unexpected Error')
 </script>
 
 <style scoped>
-.search-error {
+.unexpected-error {
   &__panel {
```
janmichek, line 42: THis should not be needed

src/errors/UnexpectedError.vue
```diff
@@ -60,8 +48,12 @@ const route = useRoute()
     }
   }
 
+  &__heading {
```
janmichek, line 51: This should not be needed

janmichek (changes requested):
Excellent, thanks for pushing things forward.
Feature-wise I am satisfied.
Found some unused code and gave some text suggestions

src/errors/SearchNotFoundError.vue
lukeromanowicz: Even better, thanks

src/errors/EntityDetailsNotFoundError.vue
lukeromanowicz: I recently did it this way and in one of the above comments Michele told me it's against the designs. I suppose we should leave Not Found here

Liubov-crypto:
I have a question: 
I see different text when I'm using search bar:

ss

 and pasting a data into url:
nf

 is this ok? 

Also there was some glitch, no text displayed on the button, maybe due to internet connection(?):
cccc

@lukeromanowicz could you check it please?

src/components/NotFoundPanel.vue
```diff
@@ -0,0 +1,63 @@
...
+
+<style scoped>
+.not-found-panel {
+  margin-top: var(--space-4);
+  padding: var(--space-4) var(--space-1);
+
+  @media (--desktop) {
+    padding: var(--space-4);
+  }
+
```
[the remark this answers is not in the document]
lukeromanowicz: Right, it was developed before the spacing update

src/errors/UnexpectedError.vue
lukeromanowicz: 100%, fixed

michele-franchi (changes requested):
Now it looks good.

One last issue, when you search for something and press the aeScan logo, the navigation bar flashes "Unexpected error". Could you check it?

https://github.com/aeternity/aescan/assets/11598532/0103740a-4c9e-475c-bb08-9c0bac20a79d

src/errors/UnexpectedError.vue
lukeromanowicz: thanks, removed

src/errors/UnexpectedError.vue
lukeromanowicz: thanks, removed

lukeromanowicz:
@michele-franchi It's not flashing unexpected error. It's flashing an empty page:
image
This is a nuxt thing that it's doing until the other pages get loaded. I can spend some more time trying to get rid of it but everything I tried so far failed.

lukeromanowicz:
@Liubov-crypto 
To answer this part `I see different text when I'm using search bar:`
Yes, that's expected by the designs

Regarding the firefox glitch - I'm not able to reproduce it in firefox. Could you, please provide the repo steps?
image

michele-franchi:
@lukeromanowicz I see an empty page but I also see "Unexpected error" in the tab title:
image

src/errors/EntityDetailsNotFoundError.vue
janmichek: ok, than. Thanks for info

janmichek (approved):
Good job, mate 🎸

Liubov-crypto:
> @Liubov-crypto To answer this part `I see different text when I'm using search bar:` Yes, that's expected by the designs
> 
> Regarding the firefox glitch - I'm not able to reproduce it in firefox. Could you, please provide the repo steps? image
> 
> Perhaps some font isn't loaded but it seems unrelated to this task

1. Enter random words into Search bar 
2. Click on Search 
3. Observe an Error

P.S. I tried to reproduce it after cleaning my cache, now works fine.

michele-franchi (commented):
I'll review again once the Unexpected Error is fixed :)

lukeromanowicz:
Thanks @michele-franchi fixed :ok_hand:

michele-franchi (approved):
It looks good now 👏 

Not sure what happened with the code, I see added files like `AccountTokensPanel` which are not part of this issue.

lukeromanowicz:
This, my good sir, was a failed rebase on my part. I'm fixing it before merging

PR #452, opened 2023-08-16
feat: Improve search bar

### LINKED ISSUE #39: Improve search bar
- If a valid hash is entered the user is redirected to the details page
- If an invalid hash is entered the user is redirected to the search results page and the specific "not found" panel is displayed. Example:

image

- If a string like "abc" is entered (not a hash), the user is redirected to the search results page to display names and tokens results

The last two cases are handled here: https://github.com/aeternity/aescan/issues/210

janmichek:
## Description
resolves #39

## Checklist:

### DIFF
--- a/src/components/SearchSelect.vue
+++ b/src/components/SearchSelect.vue
@@ -1,0 +1,88 @@
+<template>
+  <app-select
+    v-model="selectedEntity"
+    class="search-select"
+    :options="entities"
+    track-by="name"
+    label="name"
+    :hide-selected="true"
+    :searchable="false"/>
+</template>
+
+<script setup>
+import { ref } from 'vue'
+import { useVModel } from '@vueuse/core'
+import AppSelect from '@/components/AppSelect'
+
+const emit = defineEmits(['update:modelValue'])
+
+const props = defineProps({
+  modelValue: {
+    type: Object,
+    default: null,
+  },
+})
+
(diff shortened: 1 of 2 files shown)


janmichek:
I think finished the feature, but I need to wait to finish the error pages https://github.com/aeternity/aescan/pull/444

src/components/SearchSelect.vue
```diff
@@ -0,0 +1,96 @@
...
+  selectedEntity.value = preselectedEntity.value
+})
+
+const selectedEntity = useVModel(props, 'modelValue', emit)
+
+const preselectedEntity = computed(() => {
+  return entities.value.find(entity => entity.url === preselectedEntityName.value)
+})
+
+const preselectedEntityName = ref(null)
```
michele-franchi, line 41: The dropdown is not in scope for v1, as we discussed but I see it's implemented 🤔

marc0olo: yeah and this is exactly why I have been asking in the issue. it does not make sense to filter the search for all possible entities.

marc0olo: I don't even think it makes sense at all to (pre-)filter searches. a search should be convenient and if there are different result entities, it should be reflected in the search result instead

janmichek: There was a misunderstanding in the description and order in which task should come. It will be resolved diffrently

PR #458, opened 2023-08-17
feat: Improve search results page

### LINKED ISSUE #39: Improve search bar

### LINKED ISSUE #210: Improve search results page
Search results page should be improved to display names and tokens results and display not found panels for different entities.

Two tabs should be visible with names and tokens results.

Names table structure:
Name | Status (Active, In Auction, Expired)
-- | -- 

Tokens table structure:
Symbol | Name | Smart Contract ID
-- | -- | --

Token search should work on all tokens and listed ones should have the respective icon, like in the tokens page.

janmichek:
## Description
resolves #210
resolves #39

firefox_RhJxVET3LD

## Checklist:

- [x] add search page


src/components/AccountNamesTable.vue
```diff
@@ -24,7 +24,7 @@
     </thead>
     <tbody>
       <tr
-        v-for="name in accountNames?.data"
+        v-for="name in accountNames.data"
```
janmichek, line 27: added some cleanup. THe optional chaining is not needed anymore as we have loaders

src/components/AccountNamesTableCondensed.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <div class="account-names-table-condensed">
+  <div>
```
janmichek, line 2: added some cleanup. These classes are not styled

src/pages/search/[id].vue
```diff
@@ -0,0 +1,213 @@
...
+      </template>
+      <p class="search__paragraph">
+        No Accounts were found with the string <span class="search__highlighted">ak_-*{{ query }}</span>.
+      </p>
+      <p class="search__paragraph">
+        Please search using the complete <span class="search__highlighted">Account ID</span> to go directly to the
+        result.
+      </p>
+      <p class="search__paragraph">
+        Here are specific instructions as how to find it.
```
Comment (janmichek) [inline src/pages/search/[id].vue:35]: I am not super satisfied with the content, maybe we can tune it during review

src/components/SearchNamesTable.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <table class="search-names-table">
+    <tr>
```
lukeromanowicz, line 3: tbody is missing

src/components/SearchTokensTable.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+  <table class="search-tokens-table">
+    <tr>
```
lukeromanowicz, line 3: here thead/tbody are also missing

src/pages/search/[id].vue
Comment (lukeromanowicz) [reply src/pages/search/[id].vue:35]: It would say these instructions are incorrect. e.g. there is a transaction with id `th_yaMd8fYTRPEQdMudpuJ26J2Sx7RAz33t67Yg66epBzeBbocf9` but if you search for cf9 then you won't find it. Same for everything else except for tokens and names

src/pages/search/[id].vue
```diff
@@ -0,0 +1,213 @@
...
+
+    <app-panel class="search__panel">
+      <template #heading>
+        SMART CONTRACTS
+        <hint-tooltip>
+          {{ contractsHints.contract }}
+        </hint-tooltip>
+      </template>
+      <p class="search__paragraph">
+        No Smart Contracts were found with the string <span class="search__highlighted">th_-*{{ query }}</span>.
```
Comment (lukeromanowicz) [inline src/pages/search/[id].vue:66]: I would prefer to apply `q` styling just like in the error page we have discussed recently in error handling PR code review

src/components/StateChannelsTableCondensed.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="state-channels-table-condensed">
+  <div>
```
lukeromanowicz, line 2: is the div needed in such cases?

lukeromanowicz (approved):
it's working exceptionally well apart from one scenario:

When searching for 11 I am automatically redirected to block 11 while 11.chain also exist. I think in such ambiguous cases the user should be redirected to the search result page.

Very good work!

lukeromanowicz (changes requested):
picked wrong status

src/components/SearchNamesTable.vue
janmichek: ok, fixed

src/components/SearchTokensTable.vue
janmichek: fixed

src/components/StateChannelsTableCondensed.vue
janmichek: yes it is, as the class is applied from the parent component

src/pages/search/[id].vue
Comment (janmichek) [reply src/pages/search/[id].vue:35]: Can you provide some input here, please @michele-franchi

src/pages/search/[id].vue
Comment (janmichek) [reply src/pages/search/[id].vue:66]: I dunno, it still does not feel right. If it would be surrounded with quotes `""` and it had italic style I would not hesitate to use the `q`. The solution is at least controversial. Can we come up with some other ideas?

src/components/SearchTokensTable.vue
```diff
@@ -0,0 +1,79 @@
...
+    required: true,
+  },
+})
+
+</script>
+
+<style scoped>
+.search-tokens-table {
+  margin-bottom: var(--space-4);
+  /*todo move up*/
```
michele-franchi, line 66: Is this todo still valid?

src/components/SearchNamesTable.vue
```diff
@@ -0,0 +1,51 @@
...
+          </td>
+          <td>{{ name.status }}</td>
+        </tr>
+      </tbody>
+    </tbody>
+  </table>
+</template>
+
+<script setup>
+import { namesHints } from '../utils/hints/namesHints'
```
michele-franchi, line 37:
```suggestion
import { namesHints } from '@/utils/hints/namesHints'
```

src/components/SearchTokensTable.vue
```diff
@@ -0,0 +1,79 @@
...
+              class="search-tokens-table__icon"
+              :contract-id="token.contractId"/>
+            {{ token.symbol }}
+          </app-link>
+        </td>
+        <td>{{ token.name }}</td>
+        <td>
+          <value-hash-ellipsed
+            :link-to="`/contracts/${token.contractId}`"
+            :hash="token.contractId "/>
```
michele-franchi, line 43:
```suggestion
            :hash="token.contractId"/>
```

Extra space

michele-franchi (changes requested):
The code looks good but I would apply the following changes:

- All specific result panels can be hidden for now, these were part of the original idea of the dropdown. My bad for the confusion:
image

- Is it possible to reverse results? If for example I search for "test" and check names, the exact result is at the end of the list. In my opinion it should be the first one.

- Unfortunately it looks like that search is case sensitive. For example I have no results with "wetrue" but I have some results with "WeTrue". Could you please check it with mdw? In my opinion token search should work with symbol and now it just work with name

- I also agree with @lukeromanowicz that name search should be prioritised somehow. If it's hard to solve it we could even temporary disable keyblock search temporarily. Names search is more useful

src/components/SearchTokensTable.vue
janmichek: fixed

src/components/SearchNamesTable.vue
janmichek: fixed

src/components/SearchTokensTable.vue
janmichek: fixed

janmichek:
> All specific result panels can be hidden for now, these were part of the original idea of the dropdown. Let's just leave the two tabs, my bad for the confusion:

Removed

> Is it possible to reverse results? If for example I search for "test" and check names, the exact result is at the end of the list. In my opinion it should be the first one.

Will take a look on what's possible


> Unfortunately it looks like that search is case sensitive. For example I have no results with "wetrue" but I have some results with "WeTrue". Could you please check it with mdw? In my opinion token search should work with symbol and now it just work with name

Will request it from MDW


> I also agree with @lukeromanowicz that name search should be prioritised somehow. If it's hard to solve it we could even temporary disable keyblock search temporarily. Names search is more useful

Yea the only possible way is to disable it. Every keyblock number can also be a name, so logically there is no other way

lukeromanowicz:
One other way I came up with is that in such scenarios we can navigate the user to the search results page and show the name as well as a keyblock panel

janmichek:
I requested the search insensitivity here
https://github.com/aeternity/ae_mdw/issues/1549
Also added a requirement to search by token `symbol`

janmichek:
>Is it possible to reverse results? If for example I search for "test" and check names, the exact result is at the end of the list. In my opinion it should be the first one.

fixed, reversed sort order

janmichek:
Added Keyblock to search results


https://github.com/aeternity/aescan/assets/15363559/ebb9b375-8c74-40ad-b0ee-f15ee2145e94

michele-franchi (changes requested):
Personally I would not add at all the keyblock height search for now, until we find a good solution. We can just keep the one by keyblock hash.

Another thing I noticed is that If I type in the search bar a full name like "test.chain" or "test.test" I would expect to be redirected to the name page because It would be a specific one. Basically just checking if the search string ends with ".test" or ".chain". Could you add it?

lukeromanowicz (commented):
So far so good, I don't have anything to add. I'll do final review and approve when Michele's request is implemented

janmichek:
> Personally I would not add at all the keyblock height search for now, until we find a good solution. We can just keep the one by keyblock hash.
> 
> Another thing I noticed is that If I type in the search bar a full name like "test.chain" or "test.test" I would expect to be redirected to the name page because It would be a specific one. Basically just checking if the search string ends with ".test" or ".chain". Could you add it?

Search by keyblock number removed.
Search by name suffix added

https://github.com/aeternity/aescan/assets/15363559/82e34665-f611-4b09-b969-f27dfa3d9a54

src/components/TheSearchBar.vue
```diff
@@ -95,32 +80,32 @@ function isStateChannelId(query) {
 }
 
 async function isAccountName(query) {
-  return await isNameAvailable(query)
+  if (query.endsWith('.test') || query.endsWith('.chain')) {
```
michele-franchi, line 83: Not related but I would expect this function to be called simply `isName`, I don't see the correlation with account

michele-franchi (approved):
Good job 🚀
I left one not related comment.

src/components/TheSearchBar.vue
janmichek: Make sense, fixed

CharisSiarampalis:
@janmichek It seems to be working. The only issue i have is that the search appears to be case sensitive. Is this normal?

janmichek:
> @janmichek It seems to be working. The only issue i have is that the search appears to be case sensitive. Is this normal?

Thank you. Yes, we have been discussing it in this PR, and has been reported to mdw

PR #466, opened 2023-08-22
feat: transactions live sync

### LINKED ISSUE #96: Reload transaction details data when it gets synced
Right now it's necessary to manually reload the page to check if a transaction has been synced and obtain additional data.

Automatically reload transaction data once it gets synced. WS listen?

Manually reload the page.

### LINKED ISSUE #446: Loading indicator of transactions waiting for mdw sync
When data from mdw is not available we display a lot of placeholders "---" but there's nothing indicating that data will be displayed once mdw is synced.

Example:

image

lukeromanowicz:
## Description
Resolves #446 and Resolves #96

Adds loading indicator for transactions that have not been synced yet by the middleware 

## Demo
Because the MDW was heavily optimized recently, it's almost impossible to get a not synced transaction and redirect to the details page in time. Therefor I simulated this behavior in the demo.

https://github.com/aeternity/aescan/assets/46789227/6f00a38a-c9dd-4120-a7f7-283fa3fbbb19

## Checklist:


src/pages/transactions/[id].vue
```diff
@@ -44,12 +49,34 @@ const {
...
+
+  onBeforeRouteUpdate(
+    (_to, _from, next) => {
+      subscribedTransactionId.value = null
+      next()
+    },
+  )
+
+  onBeforeRouteLeave(
+    (_to, _from, next) => {
```
Comment (janmichek) [inline src/pages/transactions/[id].vue:74]: These params are not needed I guess

src/stores/webSocket.js
```diff
@@ -52,11 +71,16 @@ export const useWebSocket = defineStore('webSocket', () => {
       return
     }
     const parsedData = camelcaseKeysDeep(JSON.parse(data))
-    await processSocketMessage(parsedData)
+    if (parsedData.subscription === 'Object' && parsedData.payload?.hash === subscribedTransactionId.value) {
```
janmichek, line 74: Could this be explained via extra const binding?

janmichek (changes requested):
Looks solid! However I am not sure how to test it.
Added some minor code suggestions

src/stores/transactionDetails.js
```diff
@@ -69,11 +67,16 @@ export const useTransactionDetailsStore = defineStore('transactionDetails', () =
     transactionTypeData.value = null
   }
 
+  function processTransactionUpdate(message) {
```
janmichek, line 70: the `process` prefix it's not really clear to me what it does or what it returns. Would it make sense to name it `updateTransactionTypeData`, or similar?

src/pages/transactions/[id].vue
Comment (lukeromanowicz) [reply src/pages/transactions/[id].vue:74]: well... how else can I get to next without putting two placeholders first? :)

src/stores/webSocket.js
lukeromanowicz: I tried my best but it was really hard to come up with the name for the first part half of that if statement

src/stores/transactionDetails.js
lukeromanowicz: I named it this way to preserve consistency with the other process function but sure, I don't have any strong preferences here. Updated accordingly with your recommendation

src/stores/webSocket.js
```diff
@@ -52,11 +71,19 @@ export const useWebSocket = defineStore('webSocket', () => {
       return
     }
     const parsedData = camelcaseKeysDeep(JSON.parse(data))
-    await processSocketMessage(parsedData)
+    const isSpecificEntityMessage = parsedData.subscription === 'Object'
```
michele-franchi, line 74: Could you explain this condition? Is it correct or was it supposed to be a typeof check?

michele-franchi (approved):
It looks good to me. At the same time it looks like mdw improved sync speed so now it's basically not happening anymore but it would be still useful in case it takes longer for some cases.

src/stores/webSocket.js
lukeromanowicz: This is correct. You can subscribe to general entity type messages (e.g. Transactions, Microblocks...) or a specific entity e.g. transaction id. This is the second case. In this scenario the message.subscription is 'Object' and the payload is specific to the entity type you subscribed.

I struggled to figure out a good name for this so originally it was without extracting to a dedicated const.

src/pages/transactions/[id].vue
Comment (janmichek) [reply src/pages/transactions/[id].vue:90]: Ok, I see now.

I quess this is not gonna work?

```suggestion
    ({next}) => {
```

But the `_` prefix could be removed, right? Or why is there underscore?

janmichek (approved):
Thanks for explanation. Added one more hint, but approving in advance

lukeromanowicz:
@marc0olo we discussed that but it was hard to come up with a good design on short notice. This solution is a quick win and consistent with loading data everywhere else in the app. We'll for sure improve it once a better design is figured out and can be applied globally in the application.

PR #469, opened 2023-08-23
feat: app metrics and axios optimization

### LINKED ISSUE #453: Outgoing requests metrics
Implement measuring times of outgoing requests that can be enabled with an additional environmental variable. We need it to find out whats causing the SSR to sometimes respond very slowly.

### LINKED ISSUE #467: Fix performance issues on dev and production
### Context information
On dev and production, the application after some time gets slower. Restarting pods sometimes help. Investigate and fix the issue

lukeromanowicz:
## Description
Resolves #453 and Resolves #467 

The performance issues were caused by applying additional interceptors to the global axios object on every request

## Demo
image

The metrics cannot be seen by default unless somebody adds NUXT_PUBLIC_DEBUG_MODE to the environmental variables. Then the result will look like above

## Checklist:


janmichek (approved):
Okay, cool 👍  Thanks for investigating and coming up with the fix. I didn't know the solution would cause such performance issues. I think it was not a problem on localhost, so I didn't suspecting this

src/stores/accountDetails.js
```diff
@@ -1,6 +1,6 @@
 import { defineStore, storeToRefs } from 'pinia'
-import axios from 'axios'
 import { useRuntimeConfig } from 'nuxt/app'
+import useAxios from '@/composables/useAxios'
```
janmichek, line 3: Would it be somehow possible to destruct the variable from it, so we can save one line per store?

michele-franchi (approved):
Nicee! Blazing fast 🚀

src/stores/accountDetails.js
janmichek: I see now, thanks for explanation

PR #473, opened 2023-08-23
feat: application version in footer

### LINKED ISSUE #459: Add release notes to aescan.io
Hello,

It would be great to have the release notes page, like they are in GitHub available e.g. new changes, version aso.

Thanks,
David

lukeromanowicz:
## Description
Resolves #459

## Demo
image

'develop' will have `-dev` suffix, while prod will have no suffixes at all. 

## Checklist:


CharisSiarampalis (approved):

janmichek (approved):
Crazy cool, I like this feature a lot!


I would prefer to navigate to the particular version in the changelog (to anchor), because the latest one does not need to be deployed one yet
EG: https://github.com/aeternity/ae_mdw/blob/master/CHANGELOG.md#1570-2023-08-03 but I guess it's not possible given the link structure

nuxt.config.ts
```diff
@@ -20,6 +20,9 @@ export default defineNuxtConfig({
     apiHost: 'https://plausible.dev.service.aepps.com',
     trackLocalhost: false,
   },
+  appConfig: {
+    APP_VERSION: process.env.APP_VERSION || 'UNKNOWN',
```
michele-franchi, line 24: Can't we simply hide that if it's empty? So the `|| ' UNKNOWN'` is not needed

src/components/TheFooter.vue
```diff
@@ -40,12 +40,22 @@
         </div>
         <div class="footer__column">
           <div class="footer__network">
-            <div class="footer__version">
+            <app-link
```
michele-franchi, line 43: In my opinion I would just add the link to the version tag and not the whole text, looks a bit better to me.

src/components/TheFooter.vue
```diff
@@ -77,6 +87,11 @@ import { useStatus } from '@/stores/status'
 
 const { middlewareVersion, nodeVersion } = storeToRefs(useStatus())
 const { MIDDLEWARE_URL } = useRuntimeConfig().public
+const { APP_VERSION } = useAppConfig()
+
+const APP_CHANGELOG_URL = 'https://github.com/aeternity/aescan/blob/main/CHANGELOG.md'
```
michele-franchi, line 92:
```suggestion
const APP_RELEASES_URL = 'https://github.com/aeternity/aescan/releases'
```

I would use the releases link here

src/components/TheFooter.vue
```diff
@@ -77,6 +87,11 @@ import { useStatus } from '@/stores/status'
...
+const NODE_CHANGELOG_URL = 'https://github.com/aeternity/aeternity/releases'
+const MDW_CHANGELOG_URL = 'https://github.com/aeternity/ae_mdw/blob/master/CHANGELOG.md'
```
michele-franchi, line 94:
```suggestion
const MDW_RELEASES_URL = 'https://github.com/aeternity/ae_mdw/releases'
```

I would use the releases link here

src/components/TheFooter.vue
```diff
@@ -77,6 +87,11 @@ import { useStatus } from '@/stores/status'
 
 const { middlewareVersion, nodeVersion } = storeToRefs(useStatus())
 const { MIDDLEWARE_URL } = useRuntimeConfig().public
+const { APP_VERSION } = useAppConfig()
+
+const APP_CHANGELOG_URL = 'https://github.com/aeternity/aescan/blob/main/CHANGELOG.md'
+const NODE_CHANGELOG_URL = 'https://github.com/aeternity/aeternity/releases'
```
michele-franchi, line 93:
```suggestion
const NODE_RELEASES_URL = 'https://github.com/aeternity/aeternity/releases'
```

michele-franchi (changes requested):
I added some personal suggestions but in general it looks good!

nuxt.config.ts
lukeromanowicz: Great idea, didn't think of that :sweat_smile:

src/components/TheFooter.vue
lukeromanowicz: This won't work on on-production. Is that ok?

michele-franchi: Why wouldn't it work? It's just about to not add the link to the service name. I don't get the difference

lukeromanowicz: Sorry, minsunderstading. Of course it will :) I'll do that

lukeromanowicz:
@michele-franchi great ideas, I believe everything is fixed.

@janmichek I love that feature but I decided not to implement linking directly to a specific tag because:
1. it will not work for aescan outside production because we create releases only for released code. That means on develop and previews aescan link will navigate to github error page
2. if middleware or core team removes some release or tag the code differently than the release itself (intentionally or not) it will again navigate users to github error page. I think it's safer to link users to releases page which is always working.

I hope it makes sense

michele-franchi (approved):
Good job 👏

PR #475, opened 2023-08-24
feat: support mobile history back button

### LINKED ISSUE #337: Mobile menu should support back button to close it

lukeromanowicz:
## Description
Resolves #337 

I agree with @janmichek that this task ended up being quite a challenge.  I managed to more or less get the expected result but it isn't perfect.

## Demo

https://github.com/aeternity/aescan/assets/46789227/866b0e42-172d-4b5f-981c-fcd06d603bb1

## Checklist:


CharisSiarampalis:
@lukeromanowicz The navigation is open when going back to the previous page. Shouldn't it be close so that the user doesn't have to scroll down?

lukeromanowicz:
This is a good question. IMO if we consider the menu to be vital part of the navigation that should be supporting navigation through history button that I would say yes - it should show up when using the history buttons. Otherwise it would be incosistent

janmichek:
https://github.com/aeternity/aescan/assets/15363559/7de618ac-df27-4f52-ac16-119f2530ccb3

I think the menu should not open when going back. It should only close when going back.
Even tho it's not very logical, as a user I don't expect opening menu again when navigating back

lukeromanowicz:
@CharisSiarampalis @janmichek new version was rolled out and I think it's now working accordingly to the expectations you stated. Let me know if it's good

CharisSiarampalis (approved):

src/components/TheHeader.vue
```diff
@@ -3,7 +3,7 @@
     <div
       :class="[
         'header__container',
-        { 'header__container--open': isNavigationOpen },
+        { 'header__container--open': isNavigationDrawerOpen },
```
janmichek, line 6:
```suggestion
        { 'header__container--open': isNavigationOpen },
```

I think we can stick to original naming, or is there some consideration for renaming?

src/app.vue
```diff
@@ -97,6 +102,24 @@ if (import.meta.env.MODE !== 'production') {
     meta: [{ name: 'robots', content: 'noindex' }],
   })
 }
+
+const isHistoryBackNavigationDetected = ref(false)
```
janmichek, line 106: suggesting shorter naming, which i think it still make sense given the context

```suggestion
const isBackNavigationDetected = ref(false)
```
or
```suggestion
const isNavigateBack = ref(false)
```

janmichek (changes requested):
Bigup! 🎖️  nice solution, readable and not that much code!

I only found one missing case:
1. go to dashboard
2. open navigation
3. click browser back
4. you will be redirected to previous page (navigation is not just closed)

janmichek:
> Bigup! 🎖️ nice solution, readable and not that much code!
> 
> I only found one missing case:
> 
>     1. go to dashboard
> 
>     2. open navigation
> 
>     3. click browser back
> 
>     4. you will be redirected to previous page (navigation is not just closed)

Can you somehow imitate and pass fake history event when the navigation is open for the first time?

lukeromanowicz:
Yes, that's the only idea I got so far

src/components/TheHeader.vue
lukeromanowicz: I used NavigationDrawer just because I find it more specific and precise. Navigation is ambiguous since it may refer to bottom navigation or top navigation. Moreover, navigation on desktop is visible all the time, only the drawer is something that can be opened or closed. If this doesn't convince you I'll go back to the previous name

src/app.vue
lukeromanowicz: isBackNavigationDetected sounds better, thanks

src/components/TheHeader.vue
janmichek: I don't see a need for renaming as nothing has fundamentally changed, but if you insist on something more precise, you can use Menu instead of Drawer. Just wanted to avoid introducing new names to codebase

lukeromanowicz:
I proposed a new version that adds "#navigation-open" hash if the page was visited for the first time to get rid of scenario that Jan mentioned. The drawback is that in this very specific scenario:

- open new tab and visit aescan
- open menu
- click any of the links
- click back button twice

to go forward you will also need to press forward button twice, since the hash stays in the history. Router history does not support detecting how much history is available ahead of the user and implementing this behavior would require at least a couple of additional hours of work and testing. I think the currently proposed version is the best balance of user experience to development effort.

src/components/TheHeader.vue
lukeromanowicz: Great suggestion, I renamed it to isMobileMenuOpen

src/stores/ui.js
```diff
@@ -0,0 +1,8 @@
+import { defineStore } from 'pinia'
+export const useUiStore = defineStore('ui', () => {
```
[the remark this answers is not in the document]
lukeromanowicz: I still prefer `ui store` over `app store` because it specifically targets UI-related activity. AppStore sounds too generic to me and suggests there might be also some business logic.

janmichek:
> I proposed a new version that adds "#navigation-open" hash if the page was visited for the first time to get rid of scenario that Jan mentioned. The drawback is that in this very specific scenario:
> 
>     * open new tab and visit aescan
> 
>     * open menu
> 
>     * click any of the links
> 
>     * click back button twice
> 
> 
> to go forward you will also need to press forward button twice, since the hash stays in the history. Router history does not support detecting how much history is available ahead of the user and implementing this behavior would require at least a couple of additional hours of work and testing. I'm not even sure if it's possible to implement 100% reliably. I think the currently proposed version is the best balance of user experience to development effort.

Ok, thanks for the new solution. I would say the previous one was a one star better than this one .) What you think?

src/utils/constants.js
```diff
@@ -55,6 +55,7 @@ export const CACHE_KEY_PRICE_DATA = 'aeternity-price-data'
 export const VISIBLE_KEYBLOCKS_LIMIT = 20
 export const VISIBLE_MICROBLOCKS_LIMIT = 30
 export const VISIBLE_TRANSACTIONS_LIMIT = 4
+export const NAVIGATION_HASH = '#navigation-open'
```
janmichek, line 58: wouldn't be better to keep the same naming as the variable in the code?
```suggestion
export const MENU_HASH = '#menu-open'
```

lukeromanowicz:
@janmichek Personally, I think the latest approach is more consistent and better for the user than the previous one. It reduces big inconsistency in very particular scenario to very small issue (the need to press the button twice and only in very very specific scenario that may not happen ever to anyone)

janmichek (approved):
Good job 🍰

janmichek:
Build stopped working. Let's continue in here https://github.com/aeternity/aescan/pull/828

PR #477, opened 2023-08-25
feat: display state channel transactions count

### LINKED ISSUE #416: Show pagination info in state channel transactions
Display pagination info in state channel transactions panel. Here is how you can calculate them: https://github.com/aeternity/ae_mdw/issues/1302

lukeromanowicz:
## Description
Resolve #416 

## Demo
image

## Checklist:

### DIFF
--- a/src/components/StateChannelTransactionsPanel.vue
+++ b/src/components/StateChannelTransactionsPanel.vue
@@ -5,6 +5,7 @@
     <paginated-content
       v-model:page-index="pageIndex"
       :limit="limit"
+      :totalCount="stateChannelTxCount"
       :entities="stateChannelTransactions"
       pagination-style="history"
       @prev-clicked="loadPrevTransactions"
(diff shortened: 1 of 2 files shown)


src/components/StateChannelTransactionsPanel.vue
```diff
@@ -31,7 +32,7 @@ import { isDesktop } from '@/utils/screen'
 import PaginatedContent from '@/components/PaginatedContent'
 
 const stateChannelDetailsStore = useStateChannelDetailsStore()
-const { stateChannelTransactions } = storeToRefs(stateChannelDetailsStore)
+const { stateChannelTransactions, stateChannelTxCount } = storeToRefs(stateChannelDetailsStore)
```
janmichek, line 35: I do not really favor mixing `tx` and `transactions`. The logical question is whether it the same or a different kind of transaction. If it's the same, why is it called differently?

janmichek (changes requested):
Works nicely, just giving one code suggestion

src/components/StateChannelTransactionsPanel.vue
lukeromanowicz: That is a fair point, I copied Tx from some other function name but it looks like Transactions is the right chunk

CharisSiarampalis (approved):

janmichek (approved):
Perfecto 👌

PR #505, opened 2023-09-13
fix: nfts ordering

janmichek:
## Description
it's sorted by name now
Sorting by creation date is developed but not deployed yet. Followup https://github.com/aeternity/ae_mdw/issues/1572
As a hotfix for https://github.com/aeternity/aescan/pull/504

## Demo
image

## Checklist:

- [x] change sorting of nfts in,list

### DIFF
--- a/src/components/NftsPanel.vue
+++ b/src/components/NftsPanel.vue
@@ -41,7 +41,7 @@ async function loadNextNfts() {
 }
 
 if (process.client) {
-  await fetchNfts({ queryParameters: `/v2/aex141?limit=${limit.value}` })
+  await fetchNfts({ queryParameters: `/v2/aex141?limit=${limit.value}&direction=forward&by=name` })
 }
 
 </script>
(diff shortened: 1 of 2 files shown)


Liubov-crypto (commented):
I'm observing that sorting by name sometimes does not happen. For example, on page 2 we see NFTs with names starting with b and c, which should be on the first page. 
And also sorting NFTs with the name "test", I see it on the first page, and similar NFTs with the same name are on subsequent pages. 
@janmichek is it ok? 

https://github.com/aeternity/aescan/assets/69896204/88d3ef5b-aa56-4ac2-a29e-023ba6ee0c42

janmichek:
> I'm observing that sorting by name sometimes does not happen. For example, on page 2 we see NFTs with names starting with b and c, which should be on the first page. And also sorting NFTs with the name "test", I see it on the first page, and similar NFTs with the same name are on subsequent pages. @janmichek is it ok?
> 2023-09-13.3.50.32.mov

Thanks for reporting, absolutely not ok.
I set the bug issue for MDW https://github.com/aeternity/ae_mdw/issues/1588
@michele-franchi let us know how you want to proceed.

michele-franchi (approved):
The ordering is still not ok because we should order NFTs by creation date but at least it's better then before, I would proceed with this until mdw implements the correct ordering.

lukeromanowicz (approved):
I agree with Michele

PR #513, opened 2023-09-19
feat: link to token page for AEX-X classified tokens

### LINKED ISSUE #435: link to token page for AEX-X classified tokens
can't easily access the token page for an AEX-9 token from contract detail view
image

please provide a way to access the token page from the contract detail view in case it is classified as AEX-9 / AEX-141 (once the page exists)

not really, as token search in token view is missing for unlisted tokens. more experienced users can copy the contractId and past it into the `/tokens/` url 😄

janmichek:
## Description
resolves #435

## Demo

https://github.com/aeternity/aescan/assets/15363559/59084af4-fca3-4346-9d1a-31d996e0b869

## Checklist:

- [x] add token link to smart contract detail

### DIFF
--- a/src/components/ContractDetailsPanel.vue
+++ b/src/components/ContractDetailsPanel.vue
@@ -209,10 +209,6 @@ const props = defineProps({
   },
 })
 
-onMounted(() => {
-  console.log(props.contractDetails)
-})
-
 const contractNodeUrl = computed(() =>
   `${NODE_URL}/v3/contracts/${props.contractDetails.id}`,
 )
--- a/src/components/ContractDetailsPanel.vue
+++ b/src/components/ContractDetailsPanel.vue
@@ -30,15 +30,18 @@
             </hint-tooltip>
           </th>
           <td class="contract-details-panel__data">
-            <app-link
+            <div
               v-if="contractDetails.contractType === 'AEX-9'"
-              :to="`/tokens/${contractDetails.id}`"
-              class="contract-details-panel__token">
-              <token-symbol-icon
-                :contract-id="contractDetails.id"
-                class="contract-details-panel__icon"/>
-              {{ contractDetails.tokenDetails.symbol }}
-            </app-link>
+              class="contract-details-panel__token-container">
+              <app-link
+                :to="`/tokens/${contractDetails.id}`"
+                class="contract-details-panel__token">
+                <token-symbol-icon
+                  :contract-id="contractDetails.id"
+                  class="contract-details-panel__icon"/>
+                {{ contractDetails.tokenDetails.symbol }}
+              </app-link>
+            </div>
             <app-link
               v-if="contractDetails.contractType === 'AEX-141'"
               :to="`/nfts/${contractDetails.id}`">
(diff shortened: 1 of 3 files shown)


src/components/ContractDetailsPanel.vue
```diff
@@ -12,12 +12,40 @@
           :label="formatEllipseHash(contractDetails.id)"
           :clipboard-text="contractDetails.id"/>
       </div>
-      <app-chip v-if="contractDetails?.contractType">
+      <app-chip
+        v-if="contractDetails?.contractType"
+        size="sm">
```
janmichek, line 17: unrelated fix

src/pages/nfts/[id].vue
```diff
@@ -47,16 +47,18 @@ const hasTemplates = computed(() => {
...
-
-if (error.value) {
-  throw showError({
-    data: {
-      entityId: route.params.id,
-      entityName: 'NFT',
-    },
-    statusMessage: 'EntityDetailsNotFound',
-  })
+try {
```
Comment (janmichek) [inline src/pages/nfts/[id].vue:50]: Unrelated fix. I found that direct URL access to NFT detail page will cause infinite loading.
EG: https://aescan.dev.aepps.com/nfts/ct_2iJ5RHF2MbriceSetNiKARVii76Fr8Fi5wGY6kUHUj3wq7f3io
Could you help me to verify the solution @lukeromanowicz ?

src/components/ContractDetailsPanel.vue
```diff
@@ -180,6 +209,10 @@ const props = defineProps({
   },
 })
 
+onMounted(() => {
+  console.log(props.contractDetails)
+})
+
```
michele-franchi, line 215:
```suggestion
```

michele-franchi (changes requested):
I see some visual differences compared to other rows:
image

- There's a bigger vertical padding
- There's a right margin after the token name

Would it be possible to achieve the same row height of the other rows? Looks like we already to that for API Links?

src/components/ContractDetailsPanel.vue
janmichek: oops forgot that. Fixed

janmichek:
> I see some visual differences compared to other rows: image
> 
>     * There's a bigger vertical padding
> 
>     * There's a right margin after the token name
> 
> 
> Would it be possible to achieve the same row height of the other rows? Looks like we already to that for API Links?

Thanks for second eyes. Fixed

michele-franchi (approved):
Looks good, there's another one small issue but I approve in advance.

The entire td is clickable:
image
but ideally only the symbol and text part should be, could you adjust it?

janmichek:
Thanks fixed

> Looks good, there's another one small issue but I approve in advance.
> 
> The entire td is clickable (because of flex): image but ideally only the symbol and text part should be, could you adjust it?

Liubov-crypto (approved):
LGTM

lukeromanowicz (approved):
Looks good :+1:

src/pages/nfts/[id].vue
Comment (janmichek) [reply src/pages/nfts/[id].vue:50]: Thanks, I created the followup issue. Can you provide some hint how to improve it there? https://github.com/aeternity/aescan/issues/517 I am bit out of ideas. The 'correct' approach does not work when accessed directly. This solution it's not that effective but at least works

PR #530, opened 2023-10-16
feat: Add tab navigation

### LINKED ISSUE #497: Add tab navigation

It's not possible to directly navigate to a tab or go back to the tab that was active when navigating back.

Tabs should be handled with params in order to user browser history.

janmichek:
## Description
resolves #497

## Checklist:

- [x] add ability to navigate tabs from browser

### DIFF
--- a/src/pages/accounts/[id].vue
+++ b/src/pages/accounts/[id].vue
@@ -16,7 +16,9 @@
       :account-details="accountDetails"/>
 
     <client-only>
-      <app-tabs v-if="isTabsVisible">
+      <app-tabs
+        v-if="isTabsVisible"
+        v-model="activeTabIndex">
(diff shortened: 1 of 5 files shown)


michele-franchi (approved):
It works fine but potentially this could be done at component level.
One idea could be to add a `key` prop to `app-tab` and handle navigation directly through `app-tabs` component?

If you think it's a good idea you can create a followup for this.

janmichek:
> It works fine but potentially this could be done at component level. One idea could be to add a `key` prop to `app-tab` and handle navigation directly through `app-tabs` component?
> 
> If you think it's a good idea you can create a followup for this.

I cannot say yes or no at the moment, need some more investigation, but I will create that followup.

PR #531, opened 2023-10-17
feat: Add Balance column on Tokens Page

### LINKED ISSUE #34: Add Balance column on Tokens Page
Feature was requested from MDW. Once implemented we can use balance directly from listing response

Tracking https://github.com/aeternity/ae_mdw/issues/1256

janmichek:
## Description
resolves #34

## Demo
firefox_owDCmvSSOL

## Checklist:

- [x] display total supply

### DIFF
--- a/src/components/TokensTable.vue
+++ b/src/components/TokensTable.vue
@@ -20,6 +20,12 @@
             {{ tokensHints.smartContractId }}
           </hint-tooltip>
         </th>
+        <th>
+          Total Supply
+          <hint-tooltip>
+            {{ tokensHints.totalSupply }}
+          </hint-tooltip>
+        </th>
       </tr>
     </thead>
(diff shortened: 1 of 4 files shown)


janmichek:
I finished the task, but I am not sure about it's purposes. Depends on what we want to show

The problem is that in the listing https://mainnet.aeternity.io/mdw/v2/aex9?limit=10 there is a `event_supply` that often is 0.

I think we want to show the `total supply`. So it we show the same value in listing as well as in detail.

This `total supply` value can be obtained from node through the SDK. Which mean the value is glued to the result per every row. It's a bit heavyweight solution but it works. Pls let me know @michele-franchi

janmichek:
Reopened here https://github.com/aeternity/aescan/pull/607

PR #569, opened 2023-11-08
feat: Live update transaction details

### LINKED ISSUE #543: Live update transaction details
Transaction details should be updated live (Keyblock confirmations).

janmichek:
## Description
resolves #543

## Demo

https://github.com/aeternity/aescan/assets/15363559/e7d7268c-9942-43cd-8e26-ba99c9869a9b

## Checklist:

- [x] make confirmations count live

### DIFF
--- a/src/stores/webSocket.js
+++ b/src/stores/webSocket.js
@@ -10,20 +10,8 @@ export const useWebSocket = defineStore('webSocket', () => {
   const { updateTransactionTypeData } = useTransactionDetailsStore()
 
   const webSocket = shallowRef()
-  const isSubscribedToTransactionDetails = ref(false)
   const isSubscribedToKeyblockDetails = ref(false)
   const subscribedTransactionId = ref(null)
-
-  watch(isSubscribedToTransactionDetails, newValue => {
-    if (!webSocket.value || webSocket.value.readyState !== WebSocket.OPEN) {
-      return
-    }
-    if (newValue) {
-      webSocket.value.send('{"op":"Subscribe", "source": "node", "payload": "KeyBlocks"}')
-    } else {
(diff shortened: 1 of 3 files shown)


src/stores/webSocket.js
```diff
@@ -6,13 +6,25 @@ import { useTransactionDetailsStore } from '@/stores/transactionDetails'
...
-  const { processSocketMessage } = useRecentBlocksStore()
+  const { processSocketMessage, updateBlockHeight } = useRecentBlocksStore()
   const { updateTransactionTypeData } = useTransactionDetailsStore()
 
   const webSocket = shallowRef()
+  const isSubscribedToTransactionDetails = ref(false)
   const isSubscribedToKeyblockDetails = ref(false)
   const subscribedTransactionId = ref(null)
 
+  watch(isSubscribedToTransactionDetails, newValue => {
```
michele-franchi, line 17: Is this logic needed? Can't we just leave the subscription open?
When the webSocket is opened the subscription is already triggered:
https://github.com/aeternity/aescan/pull/569/files#diff-6eaf0a1b5ea40a2f808855772c29c950791df43f5d4e386e7e89d8cb3542589bR66
Could this cause some issues?

michele-franchi (approved):
It works good but I think some logic is over complicated. Is it really needed to close Keyblock and Microblocks subscriptions in different cases since for example it's not specific to a particular transaction?

src/stores/webSocket.js
janmichek: Yep, you are right. Can be simplified .) Thanks for the suggestion. Fixed

michele-franchi: Looks good 👏

PR #603, opened 2023-11-23
feat: Store transaction page/cursor in url of accounts page

### LINKED ISSUE #465: Refactor `notExistent` variables
The variables called `notExistent` should be named by convenction

Either `isNotExistent` or better revert logic and name it `isExistent`

yes, but not really handy naming

janmichek:
## Description
resolves #465

## Checklist:

- [ ] add condition not to reload txs data

### DIFF
--- a/src/components/TransactionsPanel.vue
+++ b/src/components/TransactionsPanel.vue
@@ -33,14 +33,15 @@ const transactionsStore = useTransactionsStore()
 const {
   transactions,
   transactionsCount,
+  isHydrated,
+  pageIndex,
 } = storeToRefs(transactionsStore)
-const { fetchTransactions, fetchTransactionsCount } = transactionsStore
+const { fetchTransactions, fetchTransactionsCount, setPageIndex } = transactionsStore
 const route = useRoute()
 const { push } = useRouter()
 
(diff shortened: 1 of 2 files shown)


Liubov-crypto:
As @janmichek described:  "for the case going back from detail to listing, the app will lose information about the current page index".

I'm getting such behaviour when I'm checking Tx details on 2nd or 3rd page and click on Go back button I will be redirected to initial page but not to the page where I was before: 

https://github.com/aeternity/aescan/assets/69896204/17e48eba-ca42-4647-9857-3b00e3800643

The rest LGTM.

janmichek:
> As @janmichek described: "for the case going back from detail to listing, the app will lose information about the current page index".
> 
> I'm getting such behaviour when I'm checking Tx details on 2nd or 3rd page and click on Go back button I will be redirected to initial page but not to the page where I was before:
> 2023-11-23.6.59.30.mov
> 
> The rest LGTM.

This has not been in the scope, but ofc we can improve it

davidyuk (approved):
Is it possible to cover this feature with a test?

janmichek:
The build stopped to work. Let's continue in here https://github.com/aeternity/aescan/pull/811

PR #549, opened 2023-10-25
feat: Wallet account view

janmichek:
## Description
fixes #536 

There are 2 minor UX issues I will try to resolve in the following (Portfolio) issue, so I don't block the review

1. After successful wallet connection there is no loader when redirecting to connected account detail
2. When clicking on Try again, the whole page gets reloaded, which is not necessary. Only scanning again should work, but for some reason it's not working https://github.com/aeternity/aescan/issues/572

## Demo

https://github.com/aeternity/aescan/assets/15363559/64887a9e-5d48-4f4d-ad37-92793c2700ff

## Checklist:

- [x] I have read and followed the [Contributing Guide](../../CONTRIBUTING.md)  
- [x] add wallet account detail

PR #604, opened 2023-11-23
fix: double labels

janmichek:
## Description
Found one regression, so I am fixing here.
It is already resolved in https://github.com/aeternity/aescan/pull/549#discussion_r1380489008 so in the case we want to resolve it earlier separately, we can merge this

## Demo
Before
image

After
image

## Checklist:

- [x] fix double labels

### DIFF
--- a/src/styles/utilities/_responsive.css
+++ b/src/styles/utilities/_responsive.css
@@ -1,12 +1,12 @@
 .u-hidden-desktop {
   @media (--desktop) {
-    display: none;
+    display: none !important;
   }
 }
 
 .u-hidden-mobile {
-  display: none;
+  display: none !important;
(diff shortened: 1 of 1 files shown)


src/styles/utilities/_responsive.css
```diff
@@ -1,12 +1,12 @@
 .u-hidden-desktop {
   @media (--desktop) {
-    display: none;
+    display: none !important;
```
janmichek, line 3: again, here the usage of !important is completely legit and serves it's intended purpose.

Liubov-crypto (approved):
Double labels has been fixed, LGTM.

michele-franchi (approved):
Looks good

PR #675, opened 2024-03-07
fix: When transaction is unavailable the app shows 500 error 

### LINKED ISSUE #674: When transaction is unavailable the app shows 500 error
### Context information

application version: 
device: 
browser: 
operating system: 

### Steps to reproduce
search for non-existetnt tx

### What is expected?
app should show 'transaction not found'

500 error

janmichek:
## Description
fixes #674

## Demo

https://github.com/aeternity/aescan/assets/15363559/3b27fa69-79c4-4723-9cad-ceb0116b99f2

## Checklist:

- [x] remove duplicate unsanitized call

### DIFF
--- a/src/pages/transactions/[id].vue
+++ b/src/pages/transactions/[id].vue
@@ -56,8 +56,6 @@ const route = useRoute()
 const { subscribedTransactionId } = storeToRefs(useWebSocket())
 const { isLoading } = useLoading()
 
-await fetchTransactionDetails(route.params.id)
-
 const isSyncing = computed(() => isLoading.value || !transactionTypeData.value)
 
 try {


michele-franchi (approved):
Looks good

Liubov-crypto:
Looks good. 

But I have a question: for example if I'm changing a little bit the Tx hash it will show me text: There is no data available, instead of 500 error.
is this expected behaviour? 


https://github.com/aeternity/aescan/assets/69896204/99b4643c-6b7c-474d-97db-94e839563edb

michele-franchi:
@Liubov-crypto It's because the th_ you entered in the search bar is not a valid tx hash. If you try on mainnet with a valid tx hash from testnet you will get the not found screen.

janmichek:
Yep this is how it was designed. It's first tested that the hash is even valid account hash.
But I agree that would make more sense from user point of view to show that the such an account has not been found event the hash is not valid account hash

PR #717, opened 2024-03-28
feat: Improve UI and readability of panels in detail pages

### LINKED ISSUE #397: Improve UI and readability of panels in detail pages
**Acceptance criteria:**
- Data columns of panels are aligned to the left
- Column description has a fixed width to make all the tables look the same size

**Note: let's make different proposals with a sample page and discuss about it**

Example reference:
image

janmichek:
## Description
Resolves #397

## Demo
image

## Checklist:

- [x] change layout


janmichek:
@michele-franchi I kicked off the visual rework. It its still rough draft. Just wanted to sync with you on the idea before. Can you check the progress, please?

michele-franchi (commented):
- Good to have the tooltips left aligned on tables
- I would go with the following structure for the hash and type:
image
This simplifies the page and removes the ugly thing to have the hash at the right side (and type for transactions) when text is aligned to the left. This option is an alternative of what you proposed regarding hash and type:
image

Of course a similar behaviour applies for all the detail pages. For example with names the header would be "Name" instead of "Hash".

- I wouldn't remove bold from table headers, I would keep the current one for now. Can be improved in a second iteration

- All tables should be aligned in the same way in a details page, for example:

image

janmichek:
Preview was rendered here:
https://pr-717-aescan.stg.service.aepps.com/

src/components/AppPanel.vue
```diff
@@ -1,13 +1,13 @@
 <template>
   <div class="panel">
     <header
-      v-if="$slots.heading"
+      v-if="$slots.title"
```
janmichek, line 4: Not related to issue. I renamed the slot names. Before was 'header' and 'heading'

Liubov-crypto (changes requested):
1. Double appearing of account: 

acc

2. is alignment here correct? 

micro

janmichek:
> 1. Double appearing of account:
> 
> 
> acc
> 
>     2. is alignment here correct?
> 
> 
> micro

fixed

Liubov-crypto (approved):
lgtm

michele-franchi (changes requested):
Really nice improvement. I would also change the following:
- Both AEX-9 and AEX-141 should be in the same row of the Smart Contract ID, for example:
image
- Mobile views should have a little bit more space, for example:
image

As you can see th and td have no space in between.

michele-franchi (changes requested):
I think something went wrong with the develop merge in the "AENS Name" details page:

image

There are different labels. This happens in all names states: active, in auction and expired with different double labels.

janmichek:
> I think something went wrong with the develop merge in the "AENS Name" details page:
> image
> 
> There are different labels. This happens in all names states: active, in auction and expired with different double labels.

Fixed, probably some failed attempt of merging

michele-franchi (approved):
Let's fix the remaining things in the followup.

PR #774, opened 2024-05-10
fix: don't enforce https if http urls in env and related fixes

davidyuk:


.env.example
```diff
@@ -6,7 +6,6 @@ NUXT_PUBLIC_NODE_URL=https://mainnet.aeternity.io
 NUXT_PUBLIC_WEBSOCKET_URL=wss://mainnet.aeternity.io/mdw/v2/websocket
 NUXT_PUBLIC_DEX_BACKEND_URL=https://dex-backend-mainnet.prd.service.aepps.com
 NUXT_PUBLIC_NETWORK_NAME=MAINNET
-NUXT_PUBLIC_NETWORK_ID=ae_mainnet
```
davidyuk, line 9: I didn't find any usages, anyway better to ask node for it

nuxt.config.ts
```diff
@@ -26,19 +26,19 @@ export default defineNuxtConfig({
...
+      NODE_URL: undefined,
+      WEBSOCKET_URL: undefined,
+      DEX_BACKEND_URL: undefined,
+      NETWORK_NAME: undefined,
+      ALTERNATIVE_NETWORK_URL: undefined,
+      ALTERNATIVE_NETWORK_NAME: undefined,
+      AE_TOKEN_ID: undefined,
+      DEBUG_MODE: undefined,
+      CONTRACT_VERIFICATION_SERVICE_URL: undefined,
+      SH_DEX_CONTRACTS: undefined,
```
davidyuk, line 41: As I understand `runtimeConfig.public` is made to define config entries and optionally provide default values. These values can be overriden by `NUXT_PUBLIC_<NAME>`.
```
SH_DEX_CONTRACTS: process.env.SH_DEX_CONTRACTS?.split(';'),
```
meaning mixing `SH_DEX_CONTRACTS` vs `NUXT_PUBLIC_SH_DEX_CONTRACTS` and in the first case it would be an array, and a string in the second.

src/composables/useAxios.js
```diff
@@ -16,7 +16,7 @@ const useAxios = () => {
       return Promise.reject(error)
     })
 
-    if (!DEBUG_MODE || DEBUG_MODE === 'false') {
+    if (!DEBUG_MODE) {
```
davidyuk, line 19: nuxt uses destr to parse env variables so `NUXT_PUBLIC_DEBUG_MODE=false` will be literally `false`

src/server/plugins/response-headers.js
```diff
@@ -1,28 +1,42 @@
...
+    'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload',
+    'Cache-control': 'no-cache',
+    'Content-Security-Policy': [
+      'default-src \'self\' *',
+      'font-src \'self\' data:',
+      'img-src \'self\' data:',
+      'script-src \'self\' \'unsafe-inline\'',
+      'style-src \'self\' \'unsafe-inline\'',
+      'frame-src \'self\'',
+      ...allowHttp ? [] : ['upgrade-insecure-requests'],
```
davidyuk, line 22: Otherwise requests to a local middleware gets upgraded and then failed because it works only over http

.env.example
janmichek: Good catch, It is not used now, however will be used for the contract read interaction feature.
I exposed the env.example earlier than it should be.

nuxt.config.ts
janmichek: The intention behind this, that some variables will be picked up from stage-specific env config
`NUXT_PUBLIC_` prefix is Nuxt aut magic

src/composables/useAxios.js
janmichek: yeah, right. Make sense

janmichek (changes requested):
Thanks for suggestions and immediate fixes.
Apart from the changes in nuxt.config.ts,
I think they are for good.

nuxt.config.ts
davidyuk: Do you have a case when you need provide variables without `NUXT_PUBLIC_`? As I see these prefixed variables would work in any case, no need to access for example `process.env.SENTRY_DSN` directly.

janmichek:
I copied the solution to our PR. https://github.com/aeternity/aescan/pull/819 So I will close this.
Apparently, the actions will not run as it is forked repository.
You can see the comments in there @michele-franchi

PR #779, opened 2024-05-11
fix: remove invisible noise in background-hero.svg

davidyuk:
As I see noise removal doesn't change anything, but it decreases svg size significantly.


janmichek (approved):
I guess that's a good change.
What tool did you used to optimize it? @davidyuk

davidyuk:
Firstly I got interested in its size while looking at the bundle in Chrome devtools. The size was suspicious because the picture is relatively simple. Then I've opened it as text and removed extra parts in xml 🙃

PR #811, opened 2024-05-23
feat: Store transaction page/cursor in url of accounts page

### LINKED ISSUE #462: Store transaction page/cursor in url of accounts page
Use case: I'm going to look though some transactions of account at https://aescan.io/accounts/<address>.

- I'm not at the first page of Transactions
- navigating to specific transaction by its hash
- after checking its details I want to go back, pressing back button
- I see account page at the tx page instead of where I was before 🙁 

The solution would be to store page/cursor in url of accounts page, like
`https://aescan.io/accounts/<address>/transactions/57491215` (cursor's value from mdw)

janmichek:
## Description
resolves #462

I choose store solution rather than project the current state to url.

It resolved the annoying UX, but it's nothing advanced. As you see in the demo, the tx data is a snapshot from the initial access to the page. EG: You go 3 pages Next, you go 3 pages back (even tho there has been new transactions created).

 Some advanced solution would require page index information from MDW. Currently, we count the page index on FE. So for the case going back from detail to listing, the app will lose information about the current page index.
I don't think such a solution is good.

I also recommend checking reference explorers, how they solve it. They have more simplistic solution

## Demo
https://github.com/aeternity/aescan/assets/15363559/a78703b6-8204-4dd5-8150-4973bb8ccbe7

## Checklist:

- [x] add condition not to reload txs data


janmichek:
Created a new PR here. It has already been approved by LIubov and Denis here https://github.com/aeternity/aescan/pull/603
Check the notes in there @michele-franchi

janmichek:
The build stopped working. Let's continue in here https://github.com/aeternity/aescan/pull/834

PR #815, opened 2024-05-28
fix: contract calls count

michele-franchi:
## Description

Smart Contract details is not working.
As a followup @janmichek, an entire page should not stop working if for example a count is not working.

## Checklist:


janmichek (approved):
Looks resolved

Issue #786, opened 2024-05-16
Simplify mobile tables


janmichek:
### Please describe the problem that your request should resolve.
I think we are doing to complicated with the tables. OTher blockchain explorers are using different and single concept.
It also relates to detail panel/table.

### Describe the solution you'd like
Leave the tables as it is on desktop view
Mobile Detail Table - 1 entry at 2 rows (currently we have 1 entry per row)
Mobile Listing (and other) Table - Horizontally scrollable if run out of space (currently weh have separate component with one line view)

janmichek:
duplicate https://github.com/aeternity/aescan/issues/954

PR #825, opened 2024-05-30
feat: UI readability followup

### LINKED ISSUE #820: UI readability followup
Followup of #397.

What can be further improved:
- [x] Transaction details panels are not aligned on mobile:
image

- [x]  Name details special pointers tooltips are not left aligned:
image

- [x] Name details special pointers are not aligned to the name details panel:
image

- [x]  On mobile token events have a typo "arguemnts":
image

- [x]  Oracle events tooltips are not left aligned:
image

- [x]  Token details page links do not have spacing:
Image

janmichek:
## Description
resolves #820 

## Checklist:

- [x] UI fixes

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -1,104 +1,113 @@
 import fs from 'fs/promises'
-import { compileTemplate } from 'vue/compiler-sfc'
-import { sentryVitePlugin } from '@sentry/vite-plugin'
+import {compileTemplate} from 'vue/compiler-sfc'
+import {sentryVitePlugin} from '@sentry/vite-plugin'
 
 // https://nuxt.com/docs/api/configuration/nuxt-config
 export default defineNuxtConfig({
-  srcDir: './src',
-  css: ['@/assets/styles/main.css'],
-  devServer: {
-    port: 8080,
-  },
-  modules: [
-    '@pinia/nuxt',
-    '@nuxtjs/plausible',
-    'nuxt-monaco-editor',
-  ],
-  imports: {
-    dirs: ['./stores'],
-  },
-  plausible: {
-    apiHost: 'https://plausible.dev.service.aepps.com',
-    trackLocalhost: false,
-  },
-  appConfig: {
-    APP_VERSION: process.env.APP_VERSION,
-  },
-  runtimeConfig: {
-    public: {
-      SENTRY_AUTH_TOKEN: process.env.SENTRY_AUTH_TOKEN,
-      SENTRY_DSN: process.env.SENTRY_DSN,
-      APP_DOMAIN: process.env.APP_DOMAIN,
-      MIDDLEWARE_URL: process.env.MIDDLEWARE_URL,
-      NODE_URL: process.env.NODE_URL,
-      WEBSOCKET_URL: process.env.WEBSOCKET_URL,
-      DEX_BACKEND_URL: process.env.DEX_BACKEND_URL,
-      NETWORK_NAME: process.env.NETWORK_NAME,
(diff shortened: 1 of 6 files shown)


michele-franchi (changes requested):
I noticed some issues are still not fixed.
1. Transaction details panels are not aligned on mobile:
image

2. Name details special pointers are not aligned on mobile:
image

janmichek:
> I noticed some issues are still not fixed.
> 
>     1. Transaction details panels are not aligned on mobile:
> 
> 
> image
> 
>     2. Name details special pointers are not aligned on mobile:
> 
> 
> image

I commented this before. I don't know what you want me to change, unless we pick different approach

janmichek:
@michele-franchi 
https://github.com/aeternity/aescan/issues/820#issuecomment-2139416301
https://github.com/aeternity/aescan/issues/820#issuecomment-2139414929

michele-franchi:
@janmichek I understand but when I look at the screen it seems there's a bug and I would expect things to be aligned the same way, for example:
image

I understand column widths are automatically calculated. Should we then maybe set a fixed height or a strategy to have them aligned?

janmichek:
> @janmichek I understand but when I look at the screen it seems there's a bug and I would expect things to be aligned the same way, for example: image
> 
> I understand column widths are automatically calculated. Should we then maybe set a fixed height or a strategy to have them aligned?

I would better solve it with different approach mentioned in here https://github.com/aeternity/aescan/issues/786

michele-franchi:
@janmichek isn't the issue you linked related to tables only? My screenshot is a panel.

janmichek:
> @janmichek isn't the issue you linked related to tables only? My screenshot is a panel.

The panel is technically table, right?
Yes initially yes. My original idea was also to extend it to detail panels/tables. I clarified my idea to the issue.
The problem partly is a big gap between mobile and desktop view.
The app breaks to mobile view pretty early when it's limited by width in my opinion. SO you can get on 1200px width the mobile view which is not optimal at all. This can be fixed by adding a third view.

Currently, I don't have any quick solution to what you mentioned. Setting the fixed height you suggesting, will have a significant effect other elements, so we should expect more detailed work with layout. Should I proceed with it?

Liubov-crypto (changes requested):
I found that some issues are still reproducible for me: 

1. Alignment in tx details: 

al

2. Is it correct alignment for Names special pointers? from the description I understood that not: 

name

3. Alignment of Data for Token events looks different on mobile: 

token events

4. Arguements is changing to Arguments on mobile: 


https://github.com/user-attachments/assets/96e07996-ccd4-47fb-afd2-a872beb8fba8
arg

arg1

5. Events for Oracles, alignments in Query details: 

oracles

janmichek:
> I found that some issues are still reproducible for me:
> 
>     1. Alignment in tx details:
> 
> 
> al
> 
>     2. Is it correct alignment for Names special pointers? from the description I understood that not:
> 
> 
> name
> 
>     3. Alignment of Data for Token events looks different on mobile:
> 
> 
> token events
> 
>     4. Arguements is changing to Arguments on mobile:
> 
> 
> https://github.com/user-attachments/assets/96e07996-ccd4-47fb-afd2-a872beb8fba8 arg
> 
> arg1
> 
>     5. Events for Oracles, alignments in Query details:
> 
> 
> oracles

Thanks for the review
ad points 1,2,5 - this is not easy to achieve. Previously in the discussion I created followup to do something conceptually different
3. the elements are laid automatically to maximize the space
4. fixed

Liubov-crypto:
ad 4 has been fixed.

Liubov-crypto (approved):
as I understood ad points 1,2,5 will be fixed separately. the rest LGTM

PR #836, opened 2024-06-10
fix: unite signing trend logic

janmichek:
## Description
I encounter a bug in develop about trend signing. It was incorrect and inconsistent.
Here I centralized the logic of trend chip used multiple times over app to the reusable component

## Demo
Before
image

after

image

## Checklist:

- [x] trend chip component

### DIFF
--- a/src/app.vue
+++ b/src/app.vue
@@ -109,6 +109,10 @@ if (import.meta.env.MODE !== 'production') {
   })
 }
 
+function logError(error) {
+  console.error(error)
+}
+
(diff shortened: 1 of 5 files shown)


src/app.vue
```diff
@@ -109,6 +109,10 @@ if (import.meta.env.MODE !== 'production') {
   })
 }
 
+function logError(error) {
```
janmichek, line 112: This block regressed with the latest merging and rebasing, so i put it back

PR #837, opened 2024-06-10
fix: Trend sign fix

janmichek:
## Description
I encounter a bug in develop about trend signing. It was incorrect and inconsistent.
Here I centralized the logic of trend chip used multiple times over app to the reusable component

## Demo
Before
image

after

image

## Checklist:

- [x] trend chip component

### DIFF
--- a/src/app.vue
+++ b/src/app.vue
@@ -109,6 +109,10 @@ if (import.meta.env.MODE !== 'production') {
   })
 }
 
+function logError(error) {
+  console.error(error)
+}
+
 const detectedHistoryNavigation = ref(null)
 
 router.options.history.listen((_to, _from, meta) => {
--- a/src/components/MarketStats.vue
+++ b/src/components/MarketStats.vue
@@ -46,7 +46,7 @@ import { useRuntimeConfig } from 'nuxt/app'
 import { MAX_AE_DISTRIBUTION } from '@/utils/constants'
 import { formatAePrice, formatNullable, formatNumber } from '@/utils/format'
 import { useMarketStatsStore } from '@/stores/marketStats'
-import TrendChip from '~/components/TrendChip'
+import TrendChip from '@/components/TrendChip'
 
 const { NETWORK_NAME } = useRuntimeConfig().public
 const selectedNetwork = `${NETWORK_NAME
--- a/src/components/MarketStats.vue
+++ b/src/components/MarketStats.vue
@@ -12,11 +12,7 @@
         </div>
         <div class="market-stats__value">
           $ {{ formatNullable(price) }}
-          <app-chip
-            class="market-stats__chip"
-            :variant="priceChipVariant">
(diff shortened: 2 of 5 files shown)


Liubov-crypto:
Are transactions missing here? 

ae

src/app.vue
```diff
@@ -109,6 +109,10 @@ if (import.meta.env.MODE !== 'production') {
   })
 }
 
+function logError(error) {
+  console.error(error)
+}
+
```
michele-franchi, line 115:
```suggestion
```
Is this used somewhere?

src/components/MarketStats.vue
```diff
@@ -50,7 +46,7 @@ import { useRuntimeConfig } from 'nuxt/app'
 import { MAX_AE_DISTRIBUTION } from '@/utils/constants'
 import { formatAePrice, formatNullable, formatNumber } from '@/utils/format'
 import { useMarketStatsStore } from '@/stores/marketStats'
-import AppChip from '@/components/AppChip'
+import TrendChip from '~/components/TrendChip'
```
michele-franchi, line 49:
```suggestion
import TrendChip from '@/components/TrendChip'
```

src/components/TrendChip.vue
```diff
@@ -0,0 +1,19 @@
...
+<script setup>
+const props = defineProps({
+  delta: {
+    type: Number,
+    required: true,
+  },
+})
+
+const priceChangeSign = computed(() => {
+  return props.delta > 0 ? '+' : ''
```
michele-franchi, line 16:
```suggestion
  if (props.delta < 0) {
    return '-'
  } else if (props.delta > 0) {
    return '+'
  } else {
    return ''
  }
```

To cover all the cases, IMO "-" should also be displayed.

src/components/TrendChip.vue
```diff
@@ -0,0 +1,19 @@
...
+})
+const priceChipVariant = computed(() => props.delta > 0 ? 'success' : 'error')
```
michele-franchi, line 18:
```suggestion
const priceChipVariant = computed(() => props.delta => 0 ? 'success' : 'error')
```

I would display 0.0 change as success, rather than error.

michele-franchi (changes requested):

src/app.vue
janmichek: yep, there https://github.com/aeternity/aescan/blob/89be1733935d8feb7dcaca41cbdd7a27f5bfd90c/src/app.vue#L65

src/components/TrendChip.vue
janmichek: But this would produce the double minus signing I just fixed. You want it this way?

michele-franchi: You are right, It would be "-" in any case.

Issue #947, opened 2024-10-14
Add N/A label for no query and response event


janmichek:
### Please describe the problem that your request should resolve.
When the content is not valid string some nonsense or nothing is printed

### Describe the solution you'd like
Print N/A label

### Is it already possible to achieve the same outcome in any other way? If so, how?
yes just sanitize the output

### Additional context
![image](https://github.com/user-attachments/assets/69a88792-d98a-4cdb-849e-9b2763d452c4)

PR #948, opened 2024-10-14
fix: Add N/A label for no query and response even

janmichek:
## Description
fixes https://github.com/aeternity/aescan/issues/947

## Demo

image

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


Liubov-crypto (approved):
LGTM. 

One more question, are these chinese symbols expected in this case ? 

query

janmichek:
> LGTM.
> 
> One more question, are these chinese symbols expected in this case ?
> 
> query

Thank you! yeah, that seems to be correct

PR #1060, opened 2025-02-03
fix: expose port 8080 in Docker

michele-franchi:
## Description
Port 8080 has to be exposed, instead of 80, because the process is started by a non-root user which can't expose ports like 80.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,9 @@ COPY --chown=node:node --from=build /app/.output ./
 USER node
 
 ENV HOST=0.0.0.0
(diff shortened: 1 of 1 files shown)


janmichek (approved):
Most tests are passing, but those affected by MDW issue does not. Thats acceptable for now, nothing we can do to fix it fast. Hopefully, I didn't overlook anything; it's sketchy currently.

I think the issue is solved. I tried new PR, based on this branch and was built successfully https://github.com/aeternity/aescan/pull/1062

Issue #909, opened 2024-09-02
Server handlers


janmichek:
### Please describe the problem that your request should resolve.
The server logic 

### Describe the solution you'd like
- [x] refactor aeCoin
- [x] refactor Oracles
- [x] refactor Keyblocks
- [x] refactor Search
- [x] refactor Microblocks
- [x] State Channels
- [x] refactor names 
- [x] simplify time formatting (UNIX time)
- [ ] refactor Charts
- [ ] refactor accounts
- [ ] refactor transactions
- [ ] handle error function
- [ ] deduplicate server logic from stores
- [ ] separate to store and API  = add one more layer between 
- [ ] remove 'raw' prefixing
- [ ] move as most logic to API to compensate sever logic
- [ ] move logic to API imitation
- [ ] move adepters to server part and serve just clean data for the app
- [ ] prepare the whole app for separate backend
- [ ] default value isDesktop and  limit
- [ ] native server get function
- [ ] move formatters to server part
- [ ] doublecheck remoed explicit imports
- [ ] request env ariables change on prod
- [ ] remove workaround [error.data.data](http://error.data.data/) || [error.data](http://error.data/)
- [ ] move count to listings
- [ ] cache
- [ ] move logic from cells
- [ ] reuse modularized API calls
- [ ] unite function naming
- [ ] revise parameters naming
- [ ] remove old adapters
- [ ] revise store default values

### Additional context
https://nuxt.com/docs/guide/directory-structure/server
utilize more server features from link above

Related issues
https://github.com/aeternity/aescan/issues/56
https://github.com/aeternity/aescan/issues/74
https://github.com/aeternity/aescan/issues/86
https://github.com/aeternity/aescan/issues/201
https://github.com/aeternity/aescan/issues/485
https://github.com/aeternity/aescan/issues/484
https://github.com/aeternity/aescan/issues/457
https://github.com/aeternity/aescan/issues/517
https://github.com/aeternity/aescan/issues/639
https://github.com/aeternity/aescan/issues/517

PR #1085, opened 2025-02-11
feat: server handlers oracles 2, 3

janmichek:
## Description

partly solves #909 

## Checklist:


janmichek:
@mmpetarpeshev here the e2e tests does not pass, but it pass on localhost. There is no problem with the app itself

PR #1103, opened 2025-03-04
fix: displaying auctions

### LINKED ISSUE #1087: Can't see a list of name auction bids
### Steps to reproduce
- open https://testnet.aescan.io/names?type=in-auction
- click on a link to https://testnet.aescan.io/names/auctiob.chain
- see Name Not Found page

### What is expected?
To see the list of bids.

I'm getting the Name Not Found page.

janmichek:
## Description
fixes #1087
partly resolve https://github.com/aeternity/aescan/issues/909

## Demo
image

## Checklist:

### DIFF
--- a/src/composables/search.js
+++ b/src/composables/search.js
@@ -1,4 +1,9 @@
+import axios from 'axios'
+import { useRuntimeConfig } from 'nuxt/app'
+
 export const useSearchStore = defineStore('search', () => {
+  const { MIDDLEWARE_URL } = useRuntimeConfig().public
+
   const namesResults = ref([])
   const tokensResults = ref([])
   const nftsResults = ref([])
@@ -28,7 +33,6 @@ export const useSearchStore = defineStore('search', () => {
     return await $fetch(`/api/search/${id}`)
   }
 
-  // todo refactor
   async function isNameAvailable(name) {
     try {
       await axios.get(`${MIDDLEWARE_URL}/names/${name}`)
--- a/src/composables/search.js
+++ b/src/composables/search.js
@@ -28,6 +28,19 @@ export const useSearchStore = defineStore('search', () => {
     return await $fetch(`/api/search/${id}`)
   }
 
+  // todo refactor
+  async function isNameAvailable(name) {
+    try {
+      await axios.get(`${MIDDLEWARE_URL}/names/${name}`)
+      return true
+    } catch (error) {
+      if (error.response.status === 404) {
+        return false
+      }
+      return null
+    }
+  }
+
   return {
     namesResults,
     tokensResults,
@@ -36,5 +49,6 @@ export const useSearchStore = defineStore('search', () => {
     fetchTokenResults,
     fetchNamesResults,
     fetchNftsResults,
+    isNameAvailable,
   }
 })
--- a/src/server/api/names/[id].js
+++ b/src/server/api/names/[id].js
@@ -9,20 +9,21 @@ const axios = useAxios()
 export default defineEventHandler(async event => {
   const id = getRouterParam(event, 'id')
 
-  const [name, block] = await Promise.all([
-    fetchaaa(id),
+  const [nameDetails, block] = await Promise.all([
+    fetchDetails(id),
     fetchKeyblocks(),
   ])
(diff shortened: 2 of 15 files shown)


janmichek:
@mmpetarpeshev There is some problem with infrasturcture. It says it needs some update

mmpetarpeshev:
> @mmpetarpeshev There is some problem with infrasturcture. It says it needs some update

PLEASE USE SLACK ,IF YOU WANT THAT TO BE SEEN QUICKLY!!!

src/composables/search.js
```diff
@@ -28,6 +33,18 @@ export const useSearchStore = defineStore('search', () => {
     return await $fetch(`/api/search/${id}`)
   }
 
+  async function isNameAvailable(name) {
```
davidyuk, line 36: Confusing naming, I understand "available" as can be claimed, but this name is already claimed. I would name it `isNameClaimed`, `isNameExist` or `isNameRegistered`

src/composables/search.js
```diff
@@ -28,6 +33,18 @@ export const useSearchStore = defineStore('search', () => {
...
+    try {
+      await axios.get(`${MIDDLEWARE_URL}/names/${name}`)
+      return true
+    } catch (error) {
+      if (error.response.status === 404) {
+        return false
+      }
+      return null
```
davidyuk, line 44: Do some logging so as not to swallow the error. I don't think `null` is useful, propose to return `false` instead

src/server/api/names/history.js
```diff
@@ -0,0 +1,35 @@
...
+  const { data } = await axios.get(url)
+
+  return adaptNameHistory(data)
+})
+
+export function adaptNameHistory(actions) {
+  const formattedData = actions.data.map(action => {
+    return {
+      type: action.type,
+      hash: action.payload.sourceTxHash || action.payload.callTxHash || action.payload.hash,
```
davidyuk, line 23: more strict version:
```
      hash: action.payload.sourceTxHash ?? action.payload.callTxHash ?? action.payload.hash,
```

src/server/api/names/[id].js
```diff
@@ -0,0 +1,136 @@
...
+    entity: 'key-blocks',
+    limit: 1,
+  })
+  const { data } = await axios.get(url)
+  return data.data[0]
+}
+
+function adaptName(name, blockHeight, blockTime) {
+  const lastBid = name?.lastBid
+  const hash = name.hash || name.lastBid.tx.nameId
```
Comment (davidyuk) [inline src/server/api/names/[id].js:66]:
```
  const hash = name.hash ?? name.lastBid.tx.nameId
```

src/server/api/names/[id].js
```diff
@@ -0,0 +1,136 @@
...
+
+  const hasRawPointers = allPointers
+    ? allPointers.some(pointer => isAddressValid(pointer.id, Encoding.Bytearray))
+    : null
+
+  return customPointers.map(pointer => {
+    return {
+      key: pointer.key,
+      pointer: hasRawPointers ? decode(pointer.id).toString() : pointer.id,
+      isRawPointer: hasRawPointers,
```
Comment (davidyuk) [inline src/server/api/names/[id].js:133]: Why are all pointers marked as raw if only one is raw?

janmichek:
@davidyuk Thanks for suggestions! Will go through them.

src/server/api/names/[id].js
Comment (janmichek) [reply src/server/api/names/[id].js:133]: its because its either all pointers are raw or none. So the condition to check only one is raw should suffice

src/composables/search.js
janmichek: The error is signalized to user thru UI. But you are right, changed to false.

src/server/api/names/[id].js
Comment (davidyuk) [reply src/server/api/names/[id].js:133]: Technically, a raw pointer doesn't interfere with any other pointers, or is this a UI restriction?

Comment (janmichek) [reply src/server/api/names/[id].js:133]: No its not UI restriction. I don't remember precisely, but it was hard to make it work, some investigation was done. I prefer to leave it like this unless proven the approach is wrong. There was no complaint so far

src/composables/search.js
janmichek: Make sense

PR #1108, opened 2025-03-11
refactor: adapters and names

davidyuk:
found while checking #1103

### DIFF
--- a/src/utils/adapters.js
+++ b/src/utils/adapters.js
@@ -1,24 +1,6 @@
 import { DateTime } from 'luxon'
 
 import { BigNumber } from 'bignumber.js'
-import { decode, Encoding, isAddressValid } from '@aeternity/aepp-sdk'
-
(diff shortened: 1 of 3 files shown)


src/utils/adapters.js
```diff
@@ -1,24 +1,6 @@
 import { DateTime } from 'luxon'
 
 import { BigNumber } from 'bignumber.js'
-import { decode, Encoding, isAddressValid } from '@aeternity/aepp-sdk'
```
janmichek, line 4: For now, as I do gradual rework, I prefer to keep the old adapters in place and remove them once the whole rework to the server handlers is done.
Adapter functions are reused in a not-intuitive way. I need to investigate a lot to make it work temporarily, hence I leave it till the end as nobrainer. But don't worry, I will take care of that soon .] It's noted here https://github.com/aeternity/aescan/issues/909

PR #1111, opened 2025-03-13
fix: Confusing scientific notation for numbers

### LINKED ISSUE #849: Confusing scientific notation for numbers
Not sure if bug or feature, but the following leads to confusions in debugging contract calls: When looking at this transaction, https://aescan.io/transactions/th_2SWUUuQZ5YS1G1ZxFQ41Dny1enTA79D98reojKm9qKB5vv1Kx1

you can see in the calldata `{ "type": "int", "value": 1e+22 }`. Especially people not from the team get the impression that this function was actually called with the param `1e+22` (we ran into this issue while debugging things around this transaction with half a dozen people). The API returns the full number, is the scientific notation intentional or maybe even happening through VueJS or some other JS autoconversion magic? I see eventual benefits of a more compact notation, but the tradeoff for this is straight confusion.

janmichek:
## Description
fixes #849

## Demo
https://pr-1111-aescan.stg.service.aepps.com/transactions/th_2SWUUuQZ5YS1G1ZxFQ41Dny1enTA79D98reojKm9qKB5vv1Kx1

image

## Checklist:

### DIFF
--- a/src/components/TransactionTypeTableContractCallTx.vue
+++ b/src/components/TransactionTypeTableContractCallTx.vue
@@ -75,7 +75,7 @@
               {{ shortArguments }}
             </div>
             <div v-show="!isCollapsed">
-              {{ formatNullable(transactionData.arguments) }}
+              {{ formatNullable(formattedArguments) }}
             </div>
             <app-button
               v-show="isArgumentsLong"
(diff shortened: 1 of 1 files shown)


janmichek:
@nikita-fuchs fixed, please help me to check .]

PR #1112, opened 2025-03-15
ci: workaround initial pipeline fail to work with second commit

mmpetarpeshev:


janmichek:
Thanks, will need to use it for a while before I can say about the result

mmpetarpeshev:
> Thanks, will need to use it for a while before I can say about the result

Please keep in mind thats not a magic :) Still the failure needs to be checked and pipeline rerun in most of the cases.

PR #1173, opened 2026-07-09
fix: prevent SSR crash when CoinGecko market stats fetch fails

venimus:
## Problem

`aescan.io` (mainnet) has been returning HTTP 500 on every page (homepage, `/blocks`, `/transactions`, etc.) in production.

Root cause traced via pod logs on the running `aescan-app` deployment: CoinGecko's public API (`GET /api/v3/coins/aeternity`, `GET /api/v3/simple/price`) is throttling requests (`429 Throttled`) — the cluster has no CoinGecko API key configured, so it's on the free/unauthenticated tier's rate limit, shared across the whole egress IP.

When that fetch fails, `fetchPrice()` in [`src/composables/marketStats.js`](https://github.com/aeternity/aescan/blob/develop/src/composables/marketStats.js) still unconditionally reads from the cache afterwards:

```js
const cachedAeternityPriceData = cache.get(CACHE_KEY_PRICE_DATA)
price.value = cachedAeternityPriceData.usd // throws if cache is still null
```

`fetchCoinStats()` has no error handling at all, so any CoinGecko failure throws directly. The confirmed crash observed in production logs was exactly:

```
TypeError: Cannot read properties of null (reading 'usd')
```

Because `memory-cache`'s cache object is a module-level singleton shared across concurrent SSR requests in the same Node process, this can escape the `initializeStores()` try/catch under load and crash the whole page render.

## Fix

- `fetchPrice()`: bail out early if the cache is still empty after a failed fetch, instead of crashing on `.usd`.
- `fetchCoinStats()`: wrap the CoinGecko call in a try/catch (matching the existing pattern in `fetchPrice()`) and bail out early if the cache is still empty, instead of throwing unguarded.

Both now degrade gracefully — the market stats widget simply doesn't render its price/market-cap figures instead of taking down the entire site.

## Note

This doesn't address the underlying rate-limiting (no CoinGecko API key configured / requests not cached long enough relative to request volume) — that's a separate, follow-up concern. This PR only fixes the crash-on-failure so a CoinGecko outage/rate-limit can no longer take the whole site down.

🤖 Generated with investigation via Copilot CLI.


venimus:
Addressed both review points in 4ca8b7e:
- Added a 1-minute failure-backoff cache marker so a CoinGecko failure no longer causes every subsequent SSR request to immediately retry (this was actually amplifying the 429 throttling in production, since the cache never got populated during an outage and every page render calls `fetchMarketStats()`).
- Split availability tracking into `isPriceAvailable`/`isMarketCapDataAvailable` and derived the public `isMarketCapAvailable` from both, so the header warning is accurate regardless of which of the two CoinGecko endpoints failed.

venimus:
Added optional CoinGecko API key support (ea27df1), per discussion — not required, purely additive:

- New server-only `COINGECKO_API_KEY` runtime config value (documented in `.env.example`), sent as the `x-cg-demo-api-key` header on both CoinGecko requests when set.
- No key is configured in any environment currently, and behavior is unchanged when it's unset — this just makes it a drop-in whenever one is created (CoinGecko's free Demo plan has much higher limits than the public tier this app currently uses).

Full lint (`eslint src`) passes with 0 errors.

